PR #1, opened 2020-04-13
Implement builder pattern for Session 

akshayknarayan:
It's useful to have various configuration options for ssh, and inconvenient to pass them all in `connect()`. Instead, implement the builder pattern for configuring an ssh connection.

Then, use the builder pattern to easily implement support for custom ssh keys (`ssh -i`).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,6 +150,7 @@ impl KnownHosts {
 pub struct SessionBuilder {
     user: Option<String>,
     port: Option<String>,
+    keyfile: Option<std::path::PathBuf>,
     known_hosts_check: KnownHosts,
 }
 
@@ -158,6 +159,7 @@ impl Default for SessionBuilder {
         Self {
             user: None,
             port: None,
+            keyfile: None,
             known_hosts_check: KnownHosts::Add,
         }
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -145,47 +145,66 @@ impl KnownHosts {
...
+            port: None,
+            keyfile: None,
+            known_hosts_check: KnownHosts::Add,
+        }
     }
+}
 
-    /// Connect to the host at the given `addr` over SSH.
+impl SessionBuilder {
+    /// Set the ssh user.
```
jonhoo, line 169: Let's also include `(ssh -l)` here.

src/lib.rs
```diff
@@ -145,47 +145,66 @@ impl KnownHosts {
...
     ///
-    /// The format of `destination` is the same as the `destination` argument to `ssh`. It may be
-    /// specified as either `[user@]hostname` or a URI of the form `ssh://[user@]hostname[:port]`.
+    /// Default `None`, which will follow your ssh config.
```
jonhoo, line 171: I find this wording a little confusing. I think we should just have it say
> Defaults to `None`.

And then the ssh config stuff is implied by the crate-level docs.

src/lib.rs
```diff
@@ -145,47 +145,66 @@ impl KnownHosts {
...
+
+    /// See [`KnownHosts`].
+    ///
+    /// Default `KnownHosts::Add`.
+    pub fn known_hosts_check(&mut self, k: KnownHosts) -> &mut Self {
+        self.known_hosts_check = k;
+        self
+    }
+
+    /// Establish the connection.
```
jonhoo, line 201: Let's copy over some of the docs from `Session::connect` to here.

jonhoo:
This looks like a great addition to me! I left some minor remarks, but if we fix those and bring up to date with `master`, I think we should be golden!

jonhoo:
Off the top of my head, we may also want to support `-o ConnectTimeout=` as an argument in the builder. I think it's decently often used in orchestration setups. Would you mind adding a `connect_timeout` method to the builder as well?

jonhoo:
Thanks!

PR #3, opened 2020-04-14
Add support for SFTP-like operations

jonhoo:


src/sftp.rs
```diff
@@ -0,0 +1,400 @@
...
+        } else {
+            self.can(Mode::Read, path)
+        }
+    }
+
+    /// Open the remote file at `path` for writing.
+    ///
+    /// If the remote file exists, it will be truncated. If it does not, it will be created.
+    ///
+    /// Note that errors may not propagate until you call [`close`], including if the remote file
```
akshayknarayan, line 215: Might be worth mentioning the `can()` check here.
Ideally, this whole module would benefit from a small doc example I think.

src/sftp.rs
```diff
@@ -0,0 +1,400 @@
...
+    // TODO: remove
+    // TODO: exists
+
+    /// Check that the given file can be opened in the given mode.
+    ///
+    /// Note that this function is potentially racy. The permissions on the server could change
+    /// between when you check and when you start a subsequent operation, causing it to fail. It
+    /// can also produce false positives, such as if you are checking whether you can write to a
+    /// file on a read-only file system that you still have write permissions on.
+    pub fn can(&mut self, mode: Mode, path: impl AsRef<std::path::Path>) -> Result<(), Error> {
```
akshayknarayan, line 68: Why `pub`?

src/sftp.rs
jonhoo: So, `can` is actually _less_ reliable than the checks that these methods now do internally, at least for writes. If you _also_ did a `can` call, you wouldn't get any more errors.

src/sftp.rs
jonhoo: In case someone wants to check permissions separately from doing an operation?

src/sftp.rs
akshayknarayan: Yeah, I think it would be good to try to give some guidance on what `can()` does vs what is provided by `write_to` internals. (i.e., when should I call each one)

src/sftp.rs
akshayknarayan: yeah, makes sense - see my comment above about maybe giving some guidance about when to use each

src/sftp.rs
jonhoo: How's ec726ba43c4d2881109b610aa7cf2a8be212e741 ?

PR #5, opened 2020-04-21
Add shell() for easy commands

jonhoo:


tests/openssh.rs
```diff
@@ -43,6 +43,32 @@ fn stdout() {
...
+
+    let child = session.shell("echo $USER").output().unwrap();
+    assert_eq!(child.stdout, b"test-user\n");
+
+    let child = session
+        .shell(r#"touch "$USER Documents""#)
+        .status()
+        .unwrap();
+    assert!(child.success());
+
```
akshayknarayan, line 59: Might be nice to add a test of pipes, since that's the main reason I would use this. Maybe:
```suggestion
    let child = session.shell(r#"echo $USER | grep -c test"#).status().unwrap(); assert!(child.success());
```

PR #6, opened 2020-04-22
First draft of async version

jonhoo:
Note that `try_wait` had to go, and `wait` has to consume `self`, since
`tokio::process` does not support those. That will change in the next
tokio release, but for now, I've just commented out the code that we'll
bring back when we can.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,7 +1,8 @@
 use super::{Error, Session};
-use std::io::prelude::*;
-use std::process::{self, Stdio};
+use std::process::Stdio;
 use tempfile::Builder;
+use tokio::io::AsyncReadExt;
+use tokio::process;
 
 /// Build a [`Session`] with options.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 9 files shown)


src/builder.rs
```diff
@@ -165,26 +166,21 @@ impl SessionBuilder {
...
-                .unwrap()
-                .read_to_string(&mut stderr)
-                .unwrap();
-
-            return Err(Error::interpret_ssh_error(&stderr));
+        let stdout = child.stdout.take().unwrap();
+        let mut stderr = child.stderr.take().unwrap();
+        let status = child.await.map_err(Error::Connect)?;
+
+        if !status.success() {
```
akshayknarayan, line 173: This is different than the old check of `if let Some(255)` right? Are the semantics with `tokio::process` different?

src/builder.rs
```diff
@@ -165,26 +166,21 @@ impl SessionBuilder {
...
+            stderr.read_to_string(&mut err).await.unwrap();
+            return Err(Error::interpret_ssh_error(&err));
         }
 
         Ok(Session {
             ctl: dir,
             addr: String::from(destination),
             terminated: false,
-            master: std::sync::Mutex::new(Some(child)),
+            master: std::sync::Mutex::new(Some((stdout, stderr))),
```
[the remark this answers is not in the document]
jonhoo: Because we've already waited for the `child`, and we know it didn't fail, so keeping it around doesn't actually make a difference. All we care about is the output channels, since it will have passed those to the backgrounded master connection process.

src/builder.rs
jonhoo: Well, normally `ssh` uses 255 for if ssh itself failed, and otherwise it just proxies the error of the remote command. When starting a master though, there is no remote command, so any error is due to ssh. While they _should_ all still be 255 (I think), it didn't seem worthwhile to check for that specifically here.

jonhoo:
Yup, that's exactly right. That's how it'd work.

PR #10, opened 2020-05-20
Escape arguments by default

jonhoo:
Since that matches `std::process::Command` more closely.
Breaks with weird remote shells, but `raw_` provides an escape hatch.

https://github.com/sfackler/shell-escape/issues/5 is a little sad.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,5 +1,6 @@
 use super::RemoteChild;
 use super::{Error, Session};
+use std::borrow::Cow;
 use std::ffi::OsStr;
 use std::io;
 use std::process::Stdio;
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -65,6 +62,21 @@
...
 //!
+//! # Remote Shells
+//!
+//! When you invoke a remote command through ssh, the remote command is executed by a shell on the
+//! remote end. That shell _interprets_ anything passed to it — it might evalute words starting
+//! with `$` as variables, split arguments by whitespace, and other things a shell is wont to do.
+//! Since that is _usually_ not what you expect to happen, `.arg("a b")` should pass a _single_
+//! argument with the value `a b`, `openssh` _escapes_ every argument (and the command itself) by
+//! default using [`shell-escape`]. This works well in most cases, but might run into issues when
+//! the remote shell has a different syntax than the shell `shell-escape` targets (bash). For
```
akshayknarayan, line 73: Maybe note that this is the remote user's login shell in most cases

src/command.rs
```diff
@@ -84,25 +82,53 @@ impl<'s> Command<'s> {
...
+    }
+
+    /// Adds multiple arguments to pass to the remote program.
+    ///
+    /// Unlike [`args`], this method does not shell-escape `args`. The arguments are passed as
+    /// written to `ssh`, which will pass them again as arguments to the remote shell. However,
+    /// since the remote shell may do argument parsing, characters such as spaces and `*` may be
+    /// interpreted by the remote shell.
+    ///
+    /// To pass a single argument see [`arg`](Command::arg).
```
akshayknarayan, line 130: This line might be out of date? I would either just remove this line, point to `raw_arg`, or both `arg` and `raw_arg`.

akshayknarayan: Actually, the `args` docs above point to `arg`, so I guess this should point to `raw_arg`.

jonhoo:
Released on `0.6.0`.

PR #12, opened 2020-06-04
Allow setting ServerAliveInterval

vitorenesduarte:
The motivation for this PR appeared here: https://github.com/jonhoo/tsunami/issues/37
The documentation is basically a copy/short version of `man ssh_config`.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -12,6 +12,7 @@ pub struct SessionBuilder {
     port: Option<String>,
     keyfile: Option<std::path::PathBuf>,
     connect_timeout: Option<String>,
+    server_alive_interval: Option<u64>,
     known_hosts_check: KnownHosts,
 }
 
@@ -22,6 +23,7 @@ impl Default for SessionBuilder {
             port: None,
             keyfile: None,
             connect_timeout: None,
+            server_alive_interval: None,
             known_hosts_check: KnownHosts::Add,
         }
     }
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -12,6 +12,8 @@ pub struct SessionBuilder {
     port: Option<String>,
     keyfile: Option<std::path::PathBuf>,
     connect_timeout: Option<String>,
+    server_alive_interval: Option<String>,
+    server_alive_count_max: Option<String>,
```
jonhoo, line 16: These can be stored as integers, since we need to use `format!` when passing them anyway.

src/builder.rs
```diff
@@ -69,6 +73,25 @@ impl SessionBuilder {
...
+    pub fn server_alive_interval(&mut self, d: std::time::Duration) -> &mut Self {
+        self.server_alive_interval = Some(d.as_secs().to_string());
+        self
+    }
+
+    /// Set the number of server alive messages (see [`server_alive_interval`]) which may be sent
+    /// without ssh receiving any messages back from the server (`ssh -o ServerAliveCountMax`).
+    ///
+    /// [`server_alive_interval`]: #method.server_alive_interval
+    pub fn server_alive_count_max(&mut self, max: usize) -> &mut Self {
```
jonhoo, line 90: I wonder how common this is. I'm inclined to only add `server_alive_interval`, and then only add `ServerAliveCountMax` if someone actually needs it (do you?).

vitorenesduarte: I don't think so. I've addressed your suggestions and force-pushed (since it's a small diff).

PR #13, opened 2020-06-05
Improve docs with note about sshd_config's MaxSessions

vitorenesduarte:


src/lib.rs
```diff
@@ -14,8 +14,10 @@
 //! [established](Session::connect), and subsequent command invocations re-use the same connection.
 //! Behind the scenes, the crate uses ssh's [`ControlMaster`] feature to multiplex the channels for
 //! the different remote commands. Because of this, each remote command is tied to the lifetime of
-//! the [`Session`] that spawned them. When the session is [closed](Session::close), the connection
-//! is severed, and there can be no outstanding remote clients.
+//! the [`Session`] that spawned them. Note that the maximum number of multiplexed remote commands
```
jonhoo, line 17: I think I'd like the sentences you added as a separate paragraph. Makes it a little easier to spot when skimming.

vitorenesduarte: Force pushed.

PR #16, opened 2020-09-01
Add an option to place the temporary path for ControlDirectory in a user-set location.

lwb4:
I'd like to be able to put the temporary directory where the control socket is created is a custom location.

This patch enables that by adding the `.control_directory()` function to the SessionBuilder.

Tried to match code style best I could, but please give suggestions for naming/style consistency as appropriate.

I tested manually, with and without the `control_directory` option. Didn't add anything to `tests/openssh.rs` as I didn't see anything similar for `keyfile`, which was my guide for this change.

If tests are requested, please give some ideas about how this should best be done -- relatively new to Rust so I may need a push in the right direction :)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,5 +1,6 @@
 use super::{Error, Session};
 use std::borrow::Cow;
+use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use tempfile::Builder;
 use tokio::io::AsyncReadExt;
@@ -10,10 +11,11 @@ use tokio::process;
 pub struct SessionBuilder {
     user: Option<String>,
     port: Option<String>,
-    keyfile: Option<std::path::PathBuf>,
+    keyfile: Option<PathBuf>,
     connect_timeout: Option<String>,
     server_alive_interval: Option<u64>,
     known_hosts_check: KnownHosts,
+    control_dir: Option<PathBuf>,
 }
 
 impl Default for SessionBuilder {
@@ -25,6 +27,7 @@ impl Default for SessionBuilder {
             connect_timeout: None,
             server_alive_interval: None,
             known_hosts_check: KnownHosts::Add,
+            control_dir: None,
         }
     }
 }
@@ -49,7 +52,7 @@ impl SessionBuilder {
     /// Set the keyfile to use (`ssh -i`).
     ///
     /// Defaults to `None`.
-    pub fn keyfile(&mut self, p: impl AsRef<std::path::Path>) -> &mut Self {
+    pub fn keyfile(&mut self, p: impl AsRef<Path>) -> &mut Self {
         self.keyfile = Some(p.as_ref().to_path_buf());
         self
     }
@@ -78,6 +81,16 @@ impl SessionBuilder {
     /// Defaults to `None`.
     pub fn server_alive_interval(&mut self, d: std::time::Duration) -> &mut Self {
         self.server_alive_interval = Some(d.as_secs());
+        self
+    }
+
+    /// Set the directory in which the temporary directory containing the control socket will
+    /// be created.
+    /// If not set, `./` will be used (the current directory).
+    ///
+    /// Defaults to `None`.
+    pub fn control_directory(&mut self, p: impl AsRef<Path>) -> &mut Self {
+        self.control_dir = Some(p.as_ref().to_path_buf());
         self
     }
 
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -81,6 +84,16 @@ impl SessionBuilder {
         self
     }
 
+    /// Set the directory in which the temporary directory containing the control socket will
+    /// be created.
```
jonhoo, line 88:
```suggestion
    /// be created.
    ///
```

src/builder.rs
```diff
@@ -81,6 +84,16 @@ impl SessionBuilder {
...
+    /// If not set, `./` will be used (the current directory).
+    ///
+    /// Defaults to `None`.
```
jonhoo, line 91: This should probably be removed.

src/builder.rs
```diff
@@ -137,9 +150,14 @@ impl SessionBuilder {
 
     pub(crate) async fn just_connect<S: AsRef<str>>(&self, host: S) -> Result<Session, Error> {
         let destination = host.as_ref();
+
+        let mut socketdir = &Path::new("./").to_path_buf();
+        if let Some(ref control_dir) = self.control_dir {
+            socketdir = control_dir;
+        }
```
jonhoo, line 157: This could probably be simplified to
```rust
let socketdir = self.control_dir.unwrap_or(&Path::new("./"));
```
I think?

jonhoo:
This looks great, thank you! It'd also fix #17 :tada: 
I left some minor notes.

I'd like to see a test, which I think should be pretty easy. Just copy the [`it_connects`](https://github.com/jonhoo/openssh-rs/blob/b3331a0f33ba69928713fe2594b312ec0aa41844/tests/openssh.rs#L14) test, modify it to set `control_dir`, and check that there is indeed a control file in the path given to `control_dir`.

lwb4:
@jonhoo Thanks! Think I addressed your comments, let me know if that covers it.

jonhoo:
Thanks! Let me take a look

tests/openssh.rs
```diff
@@ -17,6 +21,16 @@ async fn it_connects() {
     session.close().await.unwrap();
 }
 
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn control_dir() {
+    let session = SessionBuilder::default().control_directory(&path()).connect(&addr()).await.unwrap();
```
jonhoo, line 27: I think you need to run `rustfmt` on this file — that's what CI is complaining about :)

tests/openssh.rs
```diff
@@ -9,6 +9,10 @@ fn addr() -> String {
     std::env::var("TEST_HOST").unwrap_or("ssh://test-user@127.0.0.1:2222".to_string())
 }
 
+fn path() -> String {
```
jonhoo, line 12: Hmm, I think for the purposes of the test, we want to test a specific path. Something like
```rust
control_directory("control-test")
```
I don't think we need to be able to control it through an environment variable. You may need to create the control dir first, I'm not sure.

lwb4: You do indeed, that's why I left it to an env variable but I can change that approach and create the dir beforehand.

jonhoo:
You may also want to merge with master to get the CI fixes I just pushed :)

tests/openssh.rs
lwb4: Not sure it's related to my changes -- I'm seeing lots of this error in that file:
```
error[E0670]: `async fn` is not permitted in the 2015 edition
  --> /home/lincolnb/rust/openssh-rs/tests/openssh.rs:14:1
   |
14 | async fn it_connects() {
   | ^^^^^ to use `async fn`, switch to Rust 2018
   |
   = help: set `edition = "2018"` in `Cargo.toml`
   = note: for more on editions, read https://doc.rust-lang.org/edition-guide
```

jonhoo: That's weird... Do you have a very old version of Rust installed/

tests/openssh.rs
```diff
@@ -17,6 +17,21 @@ async fn it_connects() {
     session.close().await.unwrap();
 }
 
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn control_dir() {
+    let dirname = std::path::Path::new("control-test");
+    if !dirname.exists() {
```
jonhoo, line 24: I think we can probably `assert!` that this directory does not already exist. If it does, the test wouldn't actually test anything useful.

tests/openssh.rs
lwb4: No, turns out you have to specify --edition 2018 on the command line!

jonhoo (approved):

jonhoo:
Excellent, thanks for sticking with it! Published in 0.6.3 :tada:

lwb4:
Thanks for merging!

PR #24, opened 2021-07-10
Expose the ssh -F option which allows the user to select a custom ssh config file path

czakian:
I have a use case where I need to use custom ssh config files instead of the users' ~/.ssh/config file in the home directory. This change exposes the ssh -F option as documented here: https://man.openbsd.org/ssh#F. 

Thanks for making this crate and considering my pull request!
Chris.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -16,6 +16,7 @@ pub struct SessionBuilder {
     server_alive_interval: Option<u64>,
     known_hosts_check: KnownHosts,
     control_dir: Option<PathBuf>,
+    config_file: Option<PathBuf>,
 }
 
 impl Default for SessionBuilder {
@@ -28,6 +29,7 @@ impl Default for SessionBuilder {
             server_alive_interval: None,
             known_hosts_check: KnownHosts::Add,
             control_dir: None,
+            config_file: None,
         }
     }
 }
@@ -90,6 +92,16 @@ impl SessionBuilder {
     /// If not set, `./` will be used (the current directory).
     pub fn control_directory(&mut self, p: impl AsRef<Path>) -> &mut Self {
         self.control_dir = Some(p.as_ref().to_path_buf());
+        self
+    }
+
+    /// Set an alternative per-user configuration file.
+    ///
+    /// By default, ssh uses `~/.ssh/config`. This is equivalent to `ssh -F <p>`.
+    ///
+    /// Defaults to `None`.
+    pub fn config_file(&mut self, p: impl AsRef<Path>) -> &mut Self {
+        self.config_file = Some(p.as_ref().to_path_buf());
         self
     }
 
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -93,6 +95,15 @@ impl SessionBuilder {
         self
     }
 
+    /// Set an alternative per-user configuration file instead of the default ~/.ssh/config
+    /// location(`ssh -F`).
+    ///
```
jonhoo, line 100:
```suggestion
    /// Set an alternative per-user configuration file.
    ///
    /// By default, ssh uses `~/.ssh/config`. This is equivalent to `ssh -F <p>`.
    ///
```

tests/openssh.rs
```diff
@@ -35,6 +36,37 @@ async fn control_dir() {
     std::fs::remove_dir(&dirname).unwrap();
 }
 
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn config_file() {
```
jonhoo, line 41: For this test, let's check that the connection succeeds to `addr()` rather than it failing to a random localhost port. You _should_ be able to just write the value returned from `addr()` into the generated config for `config-file-test`, and then connect normally just like in the `it_connects` tests.

jonhoo (changes requested):
Thank you! Left a few smaller suggestions.

tests/openssh.rs
czakian: Updated in next commit.

czakian: looks like this test now passes with the latest update to port 2222: 

2021-07-13T00:09:01.3082187Z test config_file ... ok

Tarpaulin is still failing, but I am not sure that's caused by this change - was it failing previous to this pull request?

jonhoo: So, there is a known problem with tarpaulin on CI here, but in this case the problem is specifically that you're hard-coding `127.0.0.1`, which won't always be right on CI. You should instead use the value provided by calling `addr()` (see the other tests).

czakian: Are you proposing that I parse the addr() string like "ssh://my-user@127.0.0.1:8888" into it component parts and then write the ssh_config file like this generated from it?

   let ssh_config_contents = r#"Host config-file-test
        User my-user
        HostName 127.0.0.1
        Port 8888"#;

Or were you thinking of a one-liner replacement in the ssh_config file with a simpler to parse syntax? I'm fine with writing either, I just don't want to add parsing to the tests if that doesn't need to be added.

czakian: I took a whack at what you think you had in mind in the next revision below...regexes are hard, so I ended up writing tests for my tests...:) Open to a better option than that, but I suspect it'll get the job done.

jonhoo (approved):
Nice! That was definitely more elaborate than I thought it'd be, but I like it. Another strategy would be to have CI set the username, port, and address separately, and then have `fn addr` just use `format!` to _create_ the connection string in the format SSH expects. Rather than parsing it out. But this way works too :)

jonhoo:
Published in 0.8.1 :tada:

czakian:
Thanks for your help getting this pulled upstream!

PR #26, opened 2021-09-20
[Feature] Alternative implementation using openssh-mux-client

NobodyXu:
This PR contains alternative implementation for `Session::check` using crate `openssh-mux-client` that is only enabled when feature "enable-openssh-mux-client" is enabled.

This PR implements part of the proposal #25 .

This PR resolves #25 .

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

PR #28, opened 2021-10-23
Set default control dir to be created under `/tmp`

### LINKED ISSUE #27: Connection failure depending on the current working directory
I have a code that calls `Session::connect(...).await`. It runs fine most of the time, but fails with "failed to connect to the remote host", when run from inside a certain directory (or its children). (I.e., the only thing that changes is the current working directory.) I don't see anything special about that directory, but perhaps I'm missing something. Any pointers on what this could be, or how to go about debugging it?

NobodyXu:
Fixed #27 

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -13,6 +13,7 @@ use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::str;
 
+use dirs::home_dir;
 use once_cell::sync::OnceCell;
 use tempfile::{Builder, TempDir};
 use tokio::process;
@@ -26,8 +27,7 @@ fn get_default_control_dir<'a>() -> Result<&'a Path, Error> {
             if let Some(xdg_state_home) = env::var_os("XDG_STATE_HOME") {
                 let xdg_state_home: PathBuf = xdg_state_home.into();
                 Ok(Some(xdg_state_home.into_boxed_path()))
-            } else if let Some(home) = env::var_os("HOME") {
-                let mut path: PathBuf = home.into();
+            } else if let Some(mut path) = home_dir() {
                 path.reserve_exact(13);
                 path.push(".local");
                 path.push("state");
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -34,7 +34,7 @@ fn get_default_control_dir<'a>() -> Result<&'a Path, Error> {
         .map(|default_control_dir| {
             default_control_dir
                 .as_deref()
-                .unwrap_or_else(|| Path::new("/tmp"))
+                .unwrap_or_else(|| Path::new("./"))
         })
 }
 
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -7,13 +7,36 @@ use super::process_impl;
 use super::native_mux_impl;
 
 use std::borrow::Cow;
+use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::str;
 
+use once_cell::sync::OnceCell;
 use tempfile::{Builder, TempDir};
 use tokio::process;
+
+/// The returned `&'static Path` can be coreced to any lifetime.
+fn get_default_control_dir<'a>() -> &'a Path {
+    static DEFAULT_CONTROL_DIR: OnceCell<Box<Path>> = OnceCell::new();
+
+    DEFAULT_CONTROL_DIR.get_or_init(|| {
+        if let Some(xdg_state_home) = env::var_os("XDG_STATE_HOME") {
+            let xdg_state_home: PathBuf = xdg_state_home.into();
+            xdg_state_home.into_boxed_path()
+        } else if let Some(home) = env::var_os("HOME") {
+            let mut path: PathBuf = home.into();
+            path.reserve_exact(13);
+            path.push(".local");
+            path.push("state");
+            path.into_boxed_path()
+        } else {
+            let path: PathBuf = "/tmp".into();
+            path.into_boxed_path()
+        }
+    })
+}
 
 /// Build a [`Session`] with options.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


NobodyXu:
> I'm hesitant to do this, because it means that the control file is going to end up in a potentially globally-readable location, which might give other users on the same machine access to the connection.

I just checked the control socket created, it is created with `rw-------` so that no other user can access it.
We can also change the permission of the temporary directory we created.

> I would prefer if we instead defaulted to somewhere under `~`. Judging by the XDG Base Directory spec,

I don't think storing it in `$HOME` is a good idea.
If the program/machine terminated without cleanup, then these temporary directory would be in `$HOME` forever.

I would rather like to use `/run/user/$UID`, which resides in a tmpfs while being only accessible to the current user.



Speaking of which, I realize there is a fundamental problem with `openssh-rs`.
If the user change the uid after ssh multiplex master is spawned, then we won't be able to access that control socket.

We should probably add a warning.

jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/openssh-rs/28)

jonhoo:
Ah, sorry, I meant `$XDG_RUNTIME_DIR`, not `$XDG_STATE_HOME`. `$XDG_RUNTIME_DIR` is set to `/run/user/UID`, so we're in agreement. But we should respect the environment variable regardless for those cases where the user specifically wants to use a different location.

As for users changing UID, that I don't think is a big problem, as it should be a fairly rare occurrence.

NobodyXu:
Cool.

I would update this PR, though I would prefer #26 to be merged first.

jonhoo:
That seems reasonable. Sorry, I haven't gotten back to #26 yet — life keeps demanding time :) I promise you it's on my radar!

NobodyXu:
@jonhoo I've rebased this PR against #26 and set default control dir to `$XDG_STATE_HOME` or `$HOME/.local/state` if the former is not set, or `/tmp` if both environment variable is not set.

NobodyXu:
@jonhoo Some integrated failed due to the fact that `$HOME/.local/state` does not exist.

Shall we use `fs::create_dir_all` to create the directory, or fallback to `"/tmp"`?

jonhoo:
Let's go with `create_dir_all`. `/tmp` seems too risky to me.

NobodyXu:
> Let's go with `create_dir_all`. `/tmp` seems too risky to me.

I forgot to ask, what if `create_dir_all` also failed?

Shall we then fallback to `/tmp`, which is the only reasonable behavior I can think of?

jonhoo:
If the `create_dir_all` fails, I think we should return an error.

NobodyXu:
@jonhoo This PR currently finds home directory via looking up environment variable `$HOME`, which IMO is a bit fragile.

Shall we add something like [`dirs::state_dir`](https://docs.rs/dirs/latest/dirs/fn.state_dir.html) to the dependency?

It fallbacks to using `libc::getpwuid_r` to obtain home dir when `$HOME` is not set.

jonhoo:
Yes, that sounds like a good idea!

NobodyXu:
@jonhoo I now uses `dirs::home_dir` and I think this PR is ready to merge!

Can you please review this PR?

src/builder.rs
```diff
@@ -7,14 +7,45 @@ use super::process_impl;
...
 use tempfile::{Builder, TempDir};
 use tokio::process;
 
+/// The returned `&'static Path` can be coreced to any lifetime.
+fn get_default_control_dir<'a>() -> Result<&'a Path, Error> {
+    static DEFAULT_CONTROL_DIR: OnceCell<Option<Box<Path>>> = OnceCell::new();
+
+    DEFAULT_CONTROL_DIR
+        .get_or_try_init(|| {
+            if let Some(xdg_state_home) = env::var_os("XDG_STATE_HOME") {
```
jonhoo, line 27: Why use this instead of [`dirs::state_dir`](https://docs.rs/dirs/latest/dirs/fn.state_dir.html)?

NobodyXu: Well, I used that so that `fs::create_dir_all` is only called for `$HOME/.local/state`.

I don't know whether it is a good idea to `fs::create_all($XDG_STATE_HOME)`.

jonhoo: I don't think that should be a problem. The current process already runs as the user, so should be fine.

NobodyXu: OK, I have updated the `get_default_control_dir` to use `dirs::state_dir`.

src/builder.rs
```diff
@@ -12,9 +12,32 @@ use std::path::{Path, PathBuf};
...
+
+                Ok(Some(state_dir.into_boxed_path()))
+            } else {
+                Ok(None)
+            }
+        })
+        .map(|default_control_dir| {
+            default_control_dir
+                .as_deref()
+                .unwrap_or_else(|| Path::new("/tmp"))
```
jonhoo, line 37: I think we should _never_ fall back to `/tmp`. _That_ is a security problem waiting to happen. It should either fall back to something like `./`, or we should return an error.

NobodyXu: I have set the fallback path to `./`.

Since we uses `./` for control directory before this PR, I believe we should also use it as a fallback.

jonhoo (approved):

NobodyXu:
@jonhoo I think we can now do a major release so that users can test out the new native-mux impelemtation.

jonhoo:
Published as 0.9.0-rc.1 :tada:

NobodyXu:
Cheers!

NobodyXu:
> Published as 0.9.0-rc.1 tada

I have tried 0.9.0-rc.1 in my own binary crate [ssh-utils](https://github.com/NobodyXu/ssh-utils) and so far I haven't encountered any bug.

PR #35, opened 2022-02-04
Fix doc on stdin, stdout and stder and add changelog

### LINKED ISSUE #34: Inconsistency in crate documentation
In [crate documentation] of openssh 0.9.0-rc.1:

> And finally, our commands never default to inheriting stdin/stdout/stderr, since we expect you are using this to automate things. Instead, unless otherwise noted, all I/O ports default to [Stdio::null](https://doc.rust-lang.org/nightly/std/process/struct.Stdio.html#method.null).

This is inconsistent since now stdin/stdout/stderr in [`Command::spawn`] and [`Command::status`] is default to inherit.

And the link to `Stdio::null` is also wrong.

[crate documentation]: https://docs.rs/openssh/0.9.0-rc.1/openssh/index.html
[`Command::spawn`]: https://docs.rs/openssh/0.9.0-rc.1/openssh/struct.Command.html#method.spawn
[`Command::status`]: https://docs.rs/openssh/0.9.0-rc.1/openssh/struct.Command.html#method.status

NobodyXu:
Fixed #34 


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/openssh-rs/35)

src/lib.rs
```diff
@@ -174,6 +173,10 @@ pub use error::Error;
 mod sftp;
 pub use sftp::{Mode, RemoteFile, Sftp};
 
+#[cfg(docsrs)]
+/// Changelog for this crate, only present in doc.
+pub mod changelog;
```
jonhoo, line 178: This is an interesting way to do changelogs! I generally prefer the [keep a changelog](https://keepachangelog.com/en/1.0.0/) format though, since it's fairly standard. Could we do that in a separate PR?

src/lib.rs
```diff
@@ -33,9 +33,8 @@
 //! the remote command directly. Usually, these are the same, though not always, as highlighted in
 //! the documetantation the individual methods. See also the section below on Remote Shells.
 //!
-//! And finally, our commands never default to inheriting stdin/stdout/stderr, since we expect you
-//! are using this to automate things. Instead, unless otherwise noted, all I/O ports default to
-//! [`Stdio::null`](std::process::Stdio::null).
+//! And finally, our commands is default to inheriting stdin/stdout/stderr to be compatible with
```
jonhoo, line 36: I don't think this is actually accurate? I think we can just remove this paragraph entirely, since this is now not a difference from `std`.

jonhoo (changes requested):

src/lib.rs
NobodyXu: Sure, I have removed the module `changelog` and will create a separate PR for it.

src/lib.rs
NobodyXu: Done, I have removed this paragraph.

jonhoo (approved):

PR #37, opened 2022-02-19
feature: Add new func SessionBuilder::compression to control compression

NobodyXu:

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -149,8 +149,6 @@ impl SessionBuilder {
     ///
     /// If `~/.ssh/config` does not enable compression, then it is disabled
     /// by default.
-    ///
-    /// Defaults to `None`.
     pub fn compression(&mut self, compression: bool) -> &mut Self {
         self.compression = Some(compression);
         self
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -143,7 +143,7 @@ impl SessionBuilder {
     /// for forwarded TCP and unix-domain connections, sftp and scp
     /// connections).
     ///
-    /// NOTE that the ssh server can forcibly disable the compression.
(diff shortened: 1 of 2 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/openssh-rs/37)

src/builder.rs
```diff
@@ -137,6 +139,23 @@ impl SessionBuilder {
...
+    /// connections).
+    ///
+    /// NOTE that the ssh server can forcibly disable the compression.
+    ///
+    /// By default, ssh uses configure value set in `~/.ssh/config`.
+    ///
+    /// If `~/.ssh/config` does not enable compression, then it is disabled
+    /// by default.
+    ///
+    /// Defaults to `None`.
```
jonhoo, line 153: This last point doesn't make sense I don't think? You already say what the default is above. There is no `Option` involved here.

src/builder.rs
```diff
@@ -137,6 +139,23 @@ impl SessionBuilder {
         self
     }
 
+    /// Enable or disable compression (including stdin, stdout, stderr, data
+    /// for forwarded TCP and unix-domain connections, sftp and scp
+    /// connections).
+    ///
+    /// NOTE that the ssh server can forcibly disable the compression.
```
jonhoo, line 146:
```suggestion
    /// Note that the ssh server can forcibly disable the compression.
```

jonhoo (changes requested):
Just a few minor notes!

src/builder.rs
NobodyXu: Fixed!

jonhoo (approved):
Thanks!

PR #38, opened 2022-02-20
Move CI to GitHub Actions

jonhoo:


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/openssh-rs/38)

jonhoo:
@NobodyXu I'm working on moving our CI to GitHub Actions as it's much nicer for a bunch of things, but am running into test failures for `local_socket_forward` and `remote_socket_forward`. The [failures are](https://github.com/jonhoo/openssh-rs/runs/5266473147?check_suite_focus=true#step:7:152):
```
---- local_socket_forward stdout ----
Creating remote process
Requesting port forward
Connecting to forwarded socket
Reading
thread 'local_socket_forward' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: UnexpectedEof, error: "early eof" }', tests/openssh.rs:935:46
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- remote_socket_forward stdout ----
Requesting port forward
thread 'remote_socket_forward' panicked at 'called `Result::unwrap()` on an `Err` value: Ssh(Custom { kind: Other, error: "mux_client_forward: forwarding request failed: remote port forwarding failed for listen port 1234\r\nmuxclient: master forward request failed\r\nHost key verification failed." })', tests/openssh.rs:858:14
```
Any idea what may be causing these? I'm using the same Docker server image, so must be something different on the client, even though it's also `ubuntu-latest` just like in the old CI.

I also noticed that CI isn't running any of the `native-mux` tests, so I've enabled those now too, and hopefully they should just work :sweat_smile: I also noticed that at the moment it's hard to tell which `mux` failed in a test, since they're just run back-to-back. It may be worthwhile to add an `eprintln!` at the top of each iteration of the test loops over `connects` that prints which mux is about to be tested, so that the test output log will contain that information.

Coverage appears to still fail, but that is most likely an upstream issue (https://github.com/xd009642/tarpaulin/issues/951).

Separately, I've added a CI pass that tests [minimal dependency versions](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277), and it [fails on `openssh-mux-client-0.14.0`](https://github.com/jonhoo/openssh-rs/runs/5266473159?check_suite_focus=true#step:9:138). We probably need to bump the minimal version specified in our `Cargo.toml` to fix that.

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,51 @@
...
+          args: --all-features
+        env:
+          # makes all the ignored tests not ignored
+          RUSTFLAGS: --cfg=ci
+    services:
+      openssh:
+        image: linuxserver/openssh-server:amd64-latest
+        ports:
+          - 2222:2222
+        env:
```
NobodyXu, line 48: `linuxserver/openssh-server` does not enable ssh port forwarding by default, it requires an additional environment variable to be set:

```
DOCKER_MODS='linuxserver/mods:openssh-server-ssh-tunnel'
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,53 @@
...
+      - run: docker exec $(docker ps | grep openssh-server | awk '{print $1}') cat /config/logs/openssh/current
+        name: ssh server log
+        if: ${{ failure() }}
+    services:
+      openssh:
+        image: linuxserver/openssh-server:amd64-latest
+        ports:
+          - 2222:2222
+        env:
+          USER_NAME: test-user
```
NobodyXu, line 51: `linuxserver/openssh-server` does not enable ssh port forwarding by default, it requires an additional environment variable to be set:

```
DOCKER_MODS='linuxserver/mods:openssh-server-ssh-tunnel'
```

NobodyXu (changes requested):

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: codecov/codecov-action@v2
+        with:
+          fail_ci_if_error: true
+    services:
+      openssh:
+        image: linuxserver/openssh-server:amd64-latest
+        ports:
+          - 2222:2222
+        env:
+          USER_NAME: test-user
```
NobodyXu, line 43: `linuxserver/openssh-server` does not enable ssh port forwarding by default, it requires an additional environment variable to be set:

```
DOCKER_MODS='linuxserver/mods:openssh-server-ssh-tunnel'
```

NobodyXu (changes requested):

NobodyXu:
> Separately, I've added a CI pass that tests [minimal dependency versions](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277), and it [fails on openssh-mux-client-0.14.0](https://github.com/jonhoo/openssh-rs/runs/5266473159?check_suite_focus=true#step:9:138). We probably need to bump the minimal version specified in our Cargo.toml to fix that.

The latest release of `openssh-mux-client` is 0.14.2, so bumping it to v0.14.2 should fix the " With dependencies at minimal versions".

.github/workflows/minimal.yml
NobodyXu: > but am running into test failures for local_socket_forward and remote_socket_forward. The [failures are](https://github.com/jonhoo/openssh-rs/runs/5266473147?check_suite_focus=true#step:7:152):

Setting `DOCKER_MODS` here would fix the test failures @jonhoo

NobodyXu:
> It may be worthwhile to add an eprintln! at the top of each iteration of the test loops over connects that prints which mux is about to be tested, so that the test output log will contain that information.

Unfortunately, `connects` simply returns a `Vec` so it cannot `eprint` anything helpful.

If we can use a generator here, then we can definitely do it while avoiding the use of `Vec` altogether.

jonhoo:
> Setting `DOCKER_MODS` here would fix the test failures @jonhoo

Oh, wow, I must have somehow missed that when copying over the container definition. Thanks!

> The latest release of `openssh-mux-client` is 0.14.2, so bumping it to v0.14.2 should fix the " With dependencies at minimal versions".

Actually, this appears to be because `openssh-mux-client` requires `serde 1.0.103`, but declares that it [can use any 1.0 version](https://github.com/NobodyXu/openssh-mux-client/blob/3e1160846db74a778976d9c68d04c5e3a1872506/Cargo.toml#L16). Specifically, when trying to build `openssh-mux-client` against `serde < 1.0.103`, I get:
```
    Checking openssh-mux-client v0.14.2
error: proc-macro derive panicked
   --> /home/jon/.cargo/registry/src/github.com-1ecc6299db9ec823/openssh-mux-client-0.14.2/src/request.rs:103:45
    |
103 | #[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, TypedBuilder)]
    |                                             ^^^^^^^^^
    |
    = help: message: called `Result::unwrap()` on an `Err` value: "failed to parse derive input: \"#[builder(doc)] pub struct Session < 'a >\\n{\\n    #[builder(default = false)] pub tty : bool, #[builder(default = false)]\\n    pub x11_forwarding : bool, #[builder(default = false)] pub agent : bool,\\n    #[builder(default = false)] pub subsystem : bool,\\n    /// Set to `0xffffffff`(`char::MAX`) to disable escape character\\n    #[builder(default = char :: MAX)] pub escape_ch : char,\\n    /// Generally set to `$TERM`.\\n    #[builder(default_code =\\n              r#\\\"Cow::Borrowed(default_config::get_term().into())\\\"#)] pub term\\n    : Cow < 'a, NonZeroByteSlice >, pub cmd : Cow < 'a, NonZeroByteSlice >,\\n}\""

error[E0277]: the trait bound `Session<'_>: Serialize` is not satisfied
   --> /home/jon/.cargo/registry/src/github.com-1ecc6299db9ec823/openssh-mux-client-0.14.2/src/request.rs:85:17
    |
81  |             } => serializer.serialize_newtype_variant(
    |                             ------------------------- required by a bound introduced by this call
...
85  |                 &(*request_id, reserved, *session),
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Serialize` is not implemented for `Session<'_>`
    |
    = note: required because of the requirements on the impl of `Serialize` for `&Session<'_>`
    = note: 1 redundant requirement hidden
    = note: required because of the requirements on the impl of `Serialize` for `(u32, &&str, &Session<'_>)`
note: required by a bound in `serialize_newtype_variant`
   --> /home/jon/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.102/src/ser/mod.rs:939:12
    |
939 |         T: Serialize;
    |            ^^^^^^^^^ required by this bound in `serialize_newtype_variant`
```
That's fixed with `serde 1.0.103`, so `openssh-mux-client`'s `Cargo.toml` should be updated.

> Unfortunately, `connects` simply returns a `Vec` so it cannot `eprint` anything helpful.

It could return `Vec<(&'static str, Client)>` instead, with the first item of the tuple being the label (which we can print), no?

NobodyXu:
> It could return Vec<(&'static str, Client)> instead, with the first item of the tuple being the label (which we can print), no?

 I definitely can add that

PR #41, opened 2022-02-24
Expose a Session's control socket path

stepga:
In some situations, it's handy to have access to a SSH session's control
socket path. Therefore, expose it via a getter function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -293,7 +293,7 @@ impl Session {
     }
 
     /// Get the SSH connection's control socket path.
-    pub fn ctl(&self) -> &Path {
+    pub fn control_socket(&self) -> &Path {
         delegate!(&self.0, imp, { imp.ctl() })
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,6 +151,7 @@ compile_error!("This crate can only be used on unix");
 
 use std::borrow::Cow;
 use std::ffi::OsStr;
+use std::path::Path;
 
 mod stdio;
 pub use stdio::{ChildStderr, ChildStdin, ChildStdout, Stdio};
@@ -289,6 +290,11 @@ impl Session {
     /// Check the status of the underlying SSH connection.
     pub async fn check(&self) -> Result<(), Error> {
         delegate!(&self.0, imp, { imp.check().await })
+    }
+
+    /// Get the SSH connection's control socket path.
+    pub fn ctl(&self) -> &Path {
+        delegate!(&self.0, imp, { imp.ctl() })
     }
 
     /// Constructs a new [`Command`] for launching the program at path `program` on the remote
--- a/src/process_impl/session.rs
+++ b/src/process_impl/session.rs
@@ -188,7 +188,7 @@ impl Session {
 impl Drop for Session {
     fn drop(&mut self) {
         // Keep tempdir alive until the connection is established
-        let _ctl = match self.tempdir.take() {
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -291,6 +292,11 @@ impl Session {
         delegate!(&self.0, imp, { imp.check().await })
     }
 
+    /// Get the SSH connection's control socket path.
+    pub fn ctl_path(&self) -> Box<Path> {
```
NobodyXu, line 296: I would like ctl_path to return &Path to avoid allocation if possible.

NobodyXu (changes requested):

NobodyXu:
@stepga You would also need to change native_mux_impl/session.rs, adding this new function to it.

NobodyXu:
LGTM

stepga:
I'm sorry to have pushed yet another revision of such an easy patch, but the last revision did not compile :-/
The reason for my confusion was the different naming of both implementations' `Session` struct fields (`tempdir` & `ctl` vs `ctl` and `ctl_path`).

NobodyXu:
> I'm sorry to have pushed yet another revision of such an easy patch,

It is absolutely OK to push new revisions to fix mistakes at anytime.

> but the last revision did not compile :-/

I strongly recommend you to run script `run_ci_tests.sh` to test your implementation for bugs.

> The reason for my confusion was the different naming of both implementations' `Session` struct fields (`tempdir` & `ctl` vs `ctl` and `ctl_path`).

This change is a bit large, however I actually welcomed this change since it improves readability.

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! Scriptable SSH through OpenSSH (**only works on unix**).
+//! Scriptable SSH through OpenSSH (**only works on unix**)
```
NobodyXu, line 1: I would like the `.` to be add back here.

NobodyXu (changes requested):

src/process_impl/session.rs
```diff
@@ -184,8 +188,8 @@ impl Session {
 impl Drop for Session {
     fn drop(&mut self) {
         // Keep tempdir alive until the connection is established
-        let _ctl = match self.ctl.take() {
-            Some(ctl) => ctl,
+        let _ctl = match self.tempdir.take() {
```
NobodyXu, line 191: You can rename this variable to `_tempdir`.

NobodyXu (changes requested):

NobodyXu:
@stepga I strongly recommend you to not do force push everytime, IMHO that makes review a bit difficult.

stepga:
> @stepga I strongly recommend you to not do force push everytime, IMHO that makes review a bit difficult.

thanks for the hints.
i'm only familiar with gerrit and lurked my whole life on github without actually committing to projects.

NobodyXu:
@stepga You are welcome!

src/lib.rs
```diff
@@ -291,6 +292,11 @@ impl Session {
         delegate!(&self.0, imp, { imp.check().await })
     }
 
+    /// Get the SSH connection's control socket path.
+    pub fn ctl(&self) -> &Path {
```
jonhoo, line 296: Let's give this a more friendly name, like `control_socket`.

jonhoo (changes requested):
Thanks! Looks good to me with that small naming tweak :)

PR #50, opened 2022-03-04
Remove old "SFTP" implementation

NobodyXu:
This PR is split from #30 so that it can be merged in before release of v0.9.0 and avoid possible breaking changes required.


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/50)

src/changelog.rs
```diff
@@ -9,6 +9,13 @@ use crate::*;
 /// ## Added
 ///  - Added changelog
 ///  - Added [`SessionBuilder::compression`]
+///  - Feature `scp`.
+///  - New mod [`scp`], contains the old SFTP implementation.
+///  - Added [`Session::scp`].
+///
+/// ## Removed
+///  - `Sftp`, `Mode` and `RemoteFile`.
```
jonhoo, line 17: These weren't removed as far as I can tell?

Cargo.toml
```diff
@@ -32,6 +32,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 default = ["process-mux"]
 process-mux = []
 native-mux = ["openssh-mux-client"]
+scp = []
```
jonhoo, line 35: Hmm, why call this `scp`? That's already a separate command, but not one that this implementation uses as far as I can recall?

src/changelog.rs
NobodyXu: After mod `sftp` are renamed into `scp`, I removed these exports so now user have to type `scp::Sftp`.

Calling them removed is a little bit confusing, I will fix that.

Cargo.toml
NobodyXu: Yeah, but I couldn't figure out an alternative name though.

src/changelog.rs
NobodyXu: Done.

Cargo.toml
jonhoo: I think I'd go with something more descriptive, like `sh-file-opts`.

Alternatively we _could_ delete the module entirely. After all, There's nothing magical in there, and maybe it's better to shrink the crate's API and have users invoke `cat` and friends if they want to do file ops themselves..

NobodyXu: Well, if removing this module altogether is acceptable, then I will definitely remove it instead of keeping it around.

jonhoo: Yeah, let's do that :+1:

NobodyXu: Done!

tests/openssh.rs
```diff
@@ -376,62 +374,63 @@ macro_rules! assert_remote_kind {
 
 #[tokio::test]
 #[cfg_attr(not(ci), ignore)]
-async fn sftp_can() {
+#[cfg(feature = "scp")]
```
jonhoo, line 377: These can go away now too!

jonhoo (changes requested):

tests/openssh.rs
NobodyXu: Thanks for reminding me!

The integration tests passed, so I forgot to remove them.

jonhoo (approved):

PR #51, opened 2022-03-04
Improve local testing scripts

NobodyXu:
`run_ci_tests.sh` is refactored and new scripts `start_sshd.sh` and
`stop_sshd.sh` is created as a result.

Now, instead of starting a new sshd instance on every run to
`run_ci_tests.sh`, new sshd instance is started and shared among runs of
`run_ci_tests.sh` to speed up the testing cycle.

Another important improvements to `run_ci_tests.sh` is that now it uses
a different target directory to avoid rebuilding all dependencies.

Since IDEs usually do not set `RUSTFLAGS='--cfg=ci'`, `run_ci_tests.sh`
would usually rebuild all the dependencies and openssh.

This makes `run_ci_tests.sh` incrediably slow, and it also
affects IDEs checking, since now the IDEs also need to
rebuild the crate.

Changing the target directory to `ci-target` fixed this problem, and it
also has the new advantage that `run_ci_tests.sh` can be run
concurrently with the IDE checking and other `cargo` commands manually
run in terminal and the newly introduced `check.sh`.

`check.sh` is introduced to quickly check the syntax with flags
`--all-fatures` and run tests.

It is much faster than `run_ci_tests.sh`.

This PR is split from https://github.com/openssh-rust/openssh/pull/30 as it significantly improves the development experience.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: |
           chmod 600 .test-key
-          ssh -i .test-key -v -p 2222 -l test-user openssh -o StrictHostKeyChecking=accept-new whoami
+          ssh -i .test-key -v -p 2222 -l test-user openssh -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
         name: Test ssh connectivity
       - run: |
           eval $(ssh-agent)
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,7 +1,19 @@
 on:
   push:
     branches: [master]
+    paths-ignore:
+      - 'build_doc.sh'
+      - 'check.sh'
+      - 'run_ci_tests.sh'
+      - 'start_sshd.sh'
+      - 'stop_sshd.sh'
   pull_request:
+    paths-ignore:
+      - 'build_doc.sh'
+      - 'check.sh'
+      - 'run_ci_tests.sh'
+      - 'start_sshd.sh'
+      - 'stop_sshd.sh'
 name: coverage
 jobs:
   test:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -26,7 +26,7 @@ jobs:
       - run: |
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
-          ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
+          ssh -i .test-key -v -p 2222 -l test-user openssh -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
         name: Test ssh connectivity
       - run: |
           eval $(ssh-agent)
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -26,7 +26,7 @@ jobs:
       - run: |
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
-          ssh -i .test-key -v -p 2222 -l test-user openssh -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
+          ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
         name: Test ssh connectivity
       - run: |
           eval $(ssh-agent)
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -25,6 +25,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: |
           chmod 600 .test-key
+          mkdir /tmp/openssh-rs
           ssh -i .test-key -v -p 2222 -l test-user openssh -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
         name: Test ssh connectivity
       - run: |
--- a/check.sh
+++ b/check.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -ex
+#!/bin/bash
+
+set -euxo pipefail
 
 cd "$(dirname "$(realpath "$0")")"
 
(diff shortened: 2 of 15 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/51)

check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -ex
+
+cd $(dirname `realpath $0`)
```
jonhoo, line 3:
```suggestion
cd "$(dirname "$(realpath "$0")")"
```
To make this safe to paths with spaces in them. I highly recommend running [shellcheck](https://github.com/koalaman/shellcheck) on these scripts to catch any other latent errors.

check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -ex
```
jonhoo, line 1:
```suggestion
#!/bin/bash

set -euxo pipefail
```
`-u` ensures you don't use any unset variables, and `-o pipefail` ensures that if a command in a pipe fails, the overall command still exits with an error.

run_ci_tests.sh
```diff
@@ -2,55 +2,44 @@
...
+# Use a different target-dir since RUSTFLAGS='--cfg=ci' seems to
+# affect dependencies as well.
+#
+# Since IDEs usually do not set RUSTFLAGS='--cfg=ci', setting it
+# here would cause all the dependencies and openssh to be rebuilt.
+#
+# Thus it makes run_ci_tests.sh incrediably slow, and it also
+# affects IDEs checking, since now the IDEs also need to
+# rebuild the crate.
+CARGO_OPTS='--target-dir ci-target'
```
jonhoo, line 18:
```suggestion
export CARGO_TARGET_DIR=ci-target
```

That way you don't have to pass it around later.

run_ci_tests.sh
```diff
@@ -2,55 +2,44 @@
...
 
+function cleanup {
+    ssh-agent -k
+}
+trap cleanup EXIT
+
 echo Run tests
 rm -rf control-test config-file-test .ssh-connection*
 
+echo Remove all files in /tmp in the container
```
jonhoo, line 37: Why? Also, it's generally preferred to quote the argument to `echo`
```suggestion
echo "Remove all files in /tmp in the container"
```

start_sshd.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -ex
```
jonhoo, line 1: See various bash-ism notes from the first file.

start_sshd.sh
```diff
@@ -0,0 +1,35 @@
...
+    -e DOCKER_MODS \
+    -e PUBLIC_KEY \
+    linuxserver/openssh-server:amd64-latest
+
+export HOSTNAME=127.0.0.1
+chmod 600 .test-key
+
+# Remove 127.0.0.1:2222 from known_hosts
+ssh-keygen -R "[127.0.0.1]:2222"
+rm -f ~/.ssh/known_hosts.old
```
jonhoo, line 26: Why remove this file too?

jonhoo (changes requested):

check.sh
NobodyXu: Thanks!

I learnt something new today about bash :D

run_ci_tests.sh
NobodyXu: Sftp integration tests creates files and directories in `/tmp`.

Normally, these tests would remove any files and directories  created in `/tmp`, but in case they fail, these files will be left there and interfere with the next integration test.

start_sshd.sh
NobodyXu: `~/.ssh/known_hosts.old` sometimes cause the ssh integration test to fail.

start_sshd.sh
NobodyXu: Done.

start_sshd.sh
NobodyXu: The solution here, will be to change the path to `known_hosts` file used by the integration tests to avoid any possible interference of the user's `known_hosts` file.

NobodyXu: #54 will provide a proper solution to this problem.

run_ci_tests.sh
jonhoo: Can you add a comment about that to the file?

start_sshd.sh
jonhoo: #54 has been merged :tada:

run_ci_tests.sh
NobodyXu: Sure

.github/workflows/coverage.yml
```diff
@@ -1,7 +1,19 @@
 on:
   push:
     branches: [master]
+    paths-ignore:
```
jonhoo, line 4: This is nice! What do you think of adding a workflow that runs `shellcheck` on all `*.sh`? It should already come pre-installed in GitHub actions I think, so all you should need is a new file that contains just an `actions/checkout@v2` step + 
```yaml
 - run:
     | shellcheck *.sh
```

NobodyXu: Thanks for the advice!

We shall definitely run shellcheck on workflow to ensure the scripts are written properly.

run_ci_tests.sh
NobodyXu: Done.

start_sshd.sh
NobodyXu: Done!

.github/workflows/coverage.yml
NobodyXu: Done!

run_ci_tests.sh
```diff
@@ -1,54 +1,48 @@
...
+
+function cleanup {
+    ssh-agent -k
+}
+trap cleanup EXIT
 
 echo Run tests
 rm -rf control-test config-file-test .ssh-connection*
 
+# Sftp integration tests creates files and directories in /tmp.
```
jonhoo, line 39: Is this different now that we removed sftp?

NobodyXu: Oops, that is actually something in #30

NobodyXu: Done, it is now removed.

tests/openssh.rs
```diff
@@ -25,9 +27,25 @@ fn loopback() -> IpAddr {
     "127.0.0.1".parse().unwrap()
 }
 
-async fn session_builder_connect(builder: SessionBuilder, addr: &str) -> Vec<Session> {
+fn get_known_hosts_path() -> &'static Path {
+    lazy_static! {
```
jonhoo, line 31: Lazy static definitely seems like overkill here! It's testing code, and accessing env isn't all that expensive. So let's just make this a normal function

jonhoo (changes requested):

tests/openssh.rs
NobodyXu: Done

jonhoo (approved):

PR #52, opened 2022-03-04
Fix `Session::control_socket` for future support of windows

NobodyXu:
Return `Option<&Path>` instead of `&Path` so that we can return `None`
on windows.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -297,14 +297,10 @@ impl Session {
     }
 
     /// Get the SSH connection's control socket path.
-    ///
-    /// Currently, it will always return `Some(path)`.
-    ///
-    /// In the future, when windows support is added, it will return `None`
-    /// on Windows since openssh on Windows does not support Session
-    /// multiplex.
-    pub fn control_socket(&self) -> Option<&Path> {
-        Some(delegate!(&self.0, imp, { imp.ctl() }))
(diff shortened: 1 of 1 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/52)

jonhoo:
How do you feel about making this function  `#[cfg(not(windows))]` instead, so signal that there is no control master on windows at compile time?

NobodyXu:
> How do you feel about making this function `#[cfg(not(windows))]` instead, so signal that there is no control master on windows at compile time?

Yeah, I think using cfg here is fairly reasonable.

P.S., I am thinking of adding native ssh client implementation in the future, so that's why I propose to return an Option.

But maybe that will be something in the next breaking release, since I am still researching ssh negotiation.

jonhoo:
I explicitly don't want to implement ssh in this crate as it means re-implementing far too much of the trusted path, so I'm okay with not making this `Option` and just making it `#[cfg(not(windows))]`. :)

NobodyXu:
> I explicitly don't want to implement ssh in this crate as it means re-implementing far too much of the trusted path, so I'm okay with not making this `Option` and just making it `#[cfg(not(windows))]`. :)

Done!

jonhoo (approved):

PR #53, opened 2022-03-04
Make `ChildStd{in, out, err}` opaque types

NobodyXu:
For future Windows support.

### DIFF
--- a/src/stdio.rs
+++ b/src/stdio.rs
@@ -4,8 +4,6 @@ use super::Error;
 use super::native_mux_impl;
 
 use io_lifetimes::OwnedFd;
-use std::convert::TryFrom;
-use std::convert::TryInto;
 use std::fs::File;
 use std::io;
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
--- a/src/stdio.rs
+++ b/src/stdio.rs
@@ -4,13 +4,15 @@ use super::Error;
 use super::native_mux_impl;
 
 use io_lifetimes::OwnedFd;
+use std::convert::TryFrom;
+use std::convert::TryInto;
 use std::fs::File;
 use std::io;
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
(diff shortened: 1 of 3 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/53)

src/stdio.rs
```diff
@@ -4,13 +4,15 @@ use super::Error;
 use super::native_mux_impl;
 
 use io_lifetimes::OwnedFd;
+use std::convert::TryFrom;
+use std::convert::TryInto;
```
jonhoo, line 8: These two should no longer be necessary in the 2021 edition, which I think it's okay for us to move to as part of 0.9.0

src/stdio.rs
```diff
@@ -101,40 +103,44 @@ impl_from_for_stdio!(process::ChildStdin);
...
-pub(crate) struct ChildInputWrapper(pub(crate) ChildStdin);
-pub(crate) struct ChildOutputWrapper(pub(crate) ChildStderr);
+#[derive(Debug)]
+pub struct ChildStderr(tokio_pipe::PipeRead);
 
 macro_rules! impl_from_impl_child_io {
-    (process, $type:ident, $wrapper:ty) => {
-        impl TryFrom<tokio::process::$type> for $wrapper {
+    (process, $type:ident, $inner:ty) => {
+        impl TryFrom<tokio::process::$type> for $type {
```
jonhoo, line 143: This assumes that our wrapper will always be called the same as `tokio`'s — is that an assumption we expect will continue to hold? I guess probably?

jonhoo (commented):
Just a couple of smaller notes

src/stdio.rs
NobodyXu: I don't think we have any reason to change the name.

If we do, then it will be trivial to update this macro.

src/stdio.rs
NobodyXu: Done.

jonhoo (approved):

PR #56, opened 2022-03-05
Refactor session builder

NobodyXu:

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -8,7 +8,7 @@ use super::native_mux_impl;
 
 use std::borrow::Cow;
 use std::fs;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::process::Stdio;
 use std::str;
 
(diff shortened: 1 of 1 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/56)

jonhoo:
Hmm, why is this better?

NobodyXu:
> Hmm, why is this better?

It will make SessionBuilder much smaller.

jonhoo:
But it's a builder, and for a SSH connection at that, which is already a pretty heavy-weight thing. It's size is unlikely to matter anywhere for performance, at least as far as I can see. Do you have a use-case in mind?

NobodyXu:
> But it's a builder, and for a SSH connection at that, which is already a pretty heavy-weight thing. It's size is unlikely to matter anywhere for performance, at least as far as I can see. Do you have a use-case in mind?

Maybe for the future Windows support, where we will be connect directly without control master.

In that case, we might need to store the entire content of `SessionBuilder` in the `Session`.

jonhoo:
If we do get to that one day, I think that's a re-organization we should do then. There's no need to do it now, and it introduces some needless complexity to the code I think.

NobodyXu:
> If we do get to that one day, I think that's a re-organization we should do then. There's no need to do it now, and it introduces some needless complexity to the code I think.

OK, I will close this PR then,

PR #61, opened 2022-03-11
Add support on windows

### LINKED ISSUE #46: Unable to establish connection on Windows 11
Using version 0.8.1, creating a connection does not seem to work when on Winodws 11.

When executing the following line of code:

`let session = Session::connect(format!("root@{}", SERVER_IP.to_string()), KnownHosts::Accept).await?;`

I either get the following output, no folder in root of my project and the process terminates:

`Error: Connect(Custom { kind: ConnectionAborted, error: "getsockname failed: Bad file descriptor\r\nssh_dispatch_run_fatal: Connection to UNKNOWN port -1: unexpected internal error" })`

Or I do not get any output and the process "hangs" at this line.
The library then created a folder in the root of my project named ".ssh-connection***" before waiting indefinitely.

NobodyXu:
Fixed #46


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/61)

NobodyXu:
The Windows CI does not support running docker container, so I didn't run the integration tests on Windows CI.

jonhoo:
While I appreciate the work you've done on this, I don't think this is something I want to merge. We don't have a reliable way to test it in CI, it introduces a lot of conditional compilation, it has a completely different execution model (which will be unexpected for users), and it will become unsupported again if we eventually move to _just_ supporting the native mux.

NobodyXu:
> While I appreciate the work you've done on this, I don't think this is something I want to merge. We don't have a reliable way to test it in CI, it introduces a lot of conditional compilation, it has a completely different execution model (which will be unexpected for users), and it will become unsupported again if we eventually move to _just_ supporting the native mux.

I have to agree, as this makes this crate significantly more complicated and I haven't figure out how to run sshd on windows in the CI yet.

IMHO it will be much easier if unix socket in windows supports passing file handler/fd.

PR #62, opened 2022-03-21
PR for bumping dependencies

### LINKED ISSUE #49: 0.9.0 milestone
Since the 0.9.0-rc1 has been out for one month and there isn't any bug report for it, I think it is reasonable to prepare for a 0.9.0 release.

Tasks:
 - [x] Update changelog #57 
 - [x] Split the breaking change in #30 as a separate PR (mainly renaming existing "SFTP" implementation to `Scp`). #50
 - [x] Split the changes where improvement to local testing scripts are made as a separate PR. #51
 - [x] If windows support is to be added in the future, then we need to make `ChildStdin`, `ChildStdout` and `ChildStderr` opaque wrappers instead of concrete type and only expose trait `Unpin` (since `tokio::process::Child*` is `Unpin`), `AsyncWrite` and `AsyncRead`. (It would be much easier if rust has stablised impl type alias). #53 
 - [x] Make `Session::control_socket` available only on unix or makes it return `Option<&Path>`. It exposes implementation details and ssh on windows does not support session multiplex, so this API would not present on Windows. #52

NobodyXu:
Fixed #49 .

This PR bumps dependencies, update comment and bump the version of the crate to `v0.9.0`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ io-lifetimes = "0.6"
 once_cell = "1.10.0"
 dirs = "4.0.0"
 
-openssh-mux-client = { version = "0.14.4", optional = true }
+openssh-mux-client = { version = "0.15.0", optional = true }
 
 [dev-dependencies]
 lazy_static = "1.4.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,15 +36,15 @@ process-mux = []
 native-mux = ["openssh-mux-client"]
 
 [dependencies]
-tempfile = "3.3.0"
+tempfile = "3.1.0"
 shell-escape = "0.1.5"
 tokio = { version = "1", features = [ "process", "io-util", "macros" ] }
-tokio-pipe = "0.2.11"
+tokio-pipe = "0.2.8"
 
-libc = "0.2.121"
+libc = "0.2.112"
 io-lifetimes = "0.6"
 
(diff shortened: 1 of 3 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/62)

NobodyXu:
@jonhoo Pinging as this PR has opened for a week and it is really trivial to merge in.

jonhoo:
Sorry, time has gotten the better of me! I'd like to _not_ bump the version number yet, since we still don't know that there _won't_ be an `rc4`. Happy to do all the dependency bumps though :+1:

jonhoo:
Also, why all the minor dependency bumps? That shouldn't be necessary. The major bumps (like `io-lifetimes`) make sense though.

NobodyXu:
> Sorry, time has gotten the better of me! I'd like to _not_ bump the version number yet, since we still don't know that there _won't_ be an `rc4`.

OK, I have reverted the change, however I think we are pretty close to `0.9.0` release.

NobodyXu:
> Also, why all the minor dependency bumps? That shouldn't be necessary. The major bumps (like `io-lifetimes`) make sense though.

I indeed got too aggressive with dependency bumping and they are now reverted.

jonhoo (approved):
Thanks!

PR #66, opened 2022-04-16
Feature/support subsystem

### LINKED ISSUE #15: Sftp does not do SFTP
Instead, it is using shell commands.
This makes the tool incompatible with SFTP-only servers.

It's unclear when looking at the documentation. It probably shouldn't be called SFTP.

NobodyXu:
Support invoking ssh subsytem

Fixed #15 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,8 @@ dirs = "4.0.0"
 
 openssh-mux-client = { version = "0.15.0", optional = true }
 
+openssh-sftp-client = { version = "0.9.0", optional = true }
+
 [dev-dependencies]
 lazy_static = "1.4.0"
 regex = "1"
--- a/src/changelog.rs
+++ b/src/changelog.rs
@@ -6,6 +6,7 @@ use crate::*;
 /// ## Added
 ///  - Implement [`openssh_sftp_client::Writer`] for `ChildStdin`
 ///    if feature `openssh-sftp-client` is enabled.
+///  - [`Session::subsystem`]
 #[doc(hidden)]
 pub mod unreleased {}
 
--- a/src/changelog.rs
+++ b/src/changelog.rs
@@ -2,6 +2,10 @@
 use crate::*;
 
 /// TODO: RENAME THIS INTO THE NEXT VERSION BEFORE RELEASE
+///
+/// ## Added
+///  - Implement [`openssh_sftp_client::Writer`] for `ChildStdin`
+///    if feature `openssh-sftp-client` is enabled.
 #[doc(hidden)]
 pub mod unreleased {}
 
--- a/src/session.rs
+++ b/src/session.rs
@@ -166,6 +166,61 @@ impl Session {
         )
     }
 
+    /// Constructs a new [`Command`] for launching subsystem `program` on the remote
+    /// host.
+    ///
+    /// Unlike [`command`](Session::command), this method does not shell-escape `program`, so it may be evaluated in
+    /// unforeseen ways by the remote shell.
+    ///
+    /// The returned `Command` is a builder, with the following default configuration:
+    ///
+    /// * No arguments to the program
+    /// * Empty stdin and dsicard stdout/stderr for `spawn` or `status`, but create output pipes for
+    ///   `output`
+    ///
+    /// Builder methods are provided to change these defaults and otherwise configure the process.
+    ///
+    /// ## Sftp subsystem
+    ///
+    /// To use sftp subsystem, it is recommended to enable feature `openssh-sftp-client`,
+    /// then use the following code to construct a sftp instance:
+    ///
+    /// ```rust,no_run
+    /// # use std::error::Error;
+    /// # #[cfg(feature = "native-mux")]
+    /// # #[tokio::main]
+    /// # async fn main() -> Result<(), Box<dyn Error>> {
+    ///
+    /// use openssh::{Session, KnownHosts, Stdio};
+    /// use openssh_sftp_client::highlevel::Sftp;
+    ///
+    /// let session = Session::connect_mux("me@ssh.example.com", KnownHosts::Strict).await?;
+    ///
+    /// let mut child = session
+    ///     .subsystem("sftp")
+    ///     .stdin(Stdio::piped())
+    ///     .stdout(Stdio::piped())
+    ///     .spawn()
+    ///     .await?;
+    ///
+    /// Sftp::new(
+    ///     child.stdin().take().unwrap(),
+    ///     child.stdout().take().unwrap(),
+    ///     Default::default(),
+    /// )
+    /// .await?
+    /// .close()
+    /// .await?;
+    ///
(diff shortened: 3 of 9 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/66)

src/stdio.rs
```diff
@@ -128,6 +128,20 @@ impl_try_from_tokio_process_child_for_stdio!(ChildStderr);
 #[derive(Debug)]
 pub struct ChildStdin(tokio_pipe::PipeWrite);
 
+#[cfg(feature = "openssh-sftp-client")]
+impl openssh_sftp_client::Writer for ChildStdin {
```
jonhoo, line 132: Can we move this `impl` to `openssh_sftp_client`?

Cargo.toml
```diff
@@ -51,6 +51,8 @@ dirs = "4.0.0"
 
 openssh-mux-client = { version = "0.15.0", optional = true }
 
+openssh-sftp-client = { version = "0.9.0", optional = true }
```
jonhoo, line 54: I want to avoid bringing this in as a dependency — it _shouldn't_ be necessary, since nothing in SSH depends on SFTP (it's the other way around).

src/session.rs
```diff
@@ -166,6 +166,61 @@ impl Session {
         )
     }
 
+    /// Constructs a new [`Command`] for launching subsystem `program` on the remote
+    /// host.
+    ///
+    /// Unlike [`command`](Session::command), this method does not shell-escape `program`, so it may be evaluated in
+    /// unforeseen ways by the remote shell.
```
jonhoo, line 173: Why not?

Cargo.toml
jonhoo: It can be a `dev-dependency` so we can use it in tests/examples though!

jonhoo (changes requested):

src/stdio.rs
NobodyXu: I am afraid we can't.

The efficient implementation of this trait requires access to the underlying PipeWrite type.

src/session.rs
NobodyXu: I think in sshd, subsystem invocation are treated different from normal command, in that it may not be passed to shell at all.

For example, if the sftp-internal subsystem is enabled, then the sftp server is run directly inside sshd process.

src/stdio.rs
NobodyXu: One way to solve this is to add a new method into_inner, which directly exposes the underlying PipeWrite, but this again, exposes implementation details.

NobodyXu: Perhaps I can move this trait into another crate, such as tokio-io-utility,

jonhoo: That would work. I'm still surprised that this extra trait is even necessary. It feels like a big ergonomic hit for what seems like questionable gain over just using, say, `AsyncWrite::poll_write_vectored`.

NobodyXu: I could remove this requirement by removing the atomic write support in `WriteEnd`.

Perhaps I really should think of an alternative design for `openssh_sftp_client` to make its internal more simpler.

jonhoo: I think that might be valuable. In general it feels excessive for the protocol to require that consumers implement as low-level a trait as `AsyncWrite` themselves in a custom way.

NobodyXu: > I think that might be valuable. In general it feels excessive for the protocol to require that consumers implement as low-level a trait as `AsyncWrite` themselves in a custom way.

Yeah, it is a bit excessive, though most users probably wouldn't notice as they just pass PipeWrite to it.

NobodyXu: @jonhoo It's done!

[openssh-sftp-client-0.10.0](https://docs.rs/openssh-sftp-client/0.10.0/openssh_sftp_client/) is released with the refactor done, so we no longer has to depend on `openssh-sftp-client`.

Cargo.toml
NobodyXu: Done!

src/changelog.rs
```diff
@@ -2,6 +2,11 @@
 use crate::*;
 
 /// TODO: RENAME THIS INTO THE NEXT VERSION BEFORE RELEASE
+///
+/// ## Added
+///  - Implement [`openssh_sftp_client::Writer`] for `ChildStdin`
```
jonhoo, line 7: This is no longer accurate.

src/session.rs
```diff
@@ -166,6 +166,61 @@ impl Session {
...
+    ///
+    /// * No arguments to the program
+    /// * Empty stdin and dsicard stdout/stderr for `spawn` or `status`, but create output pipes for
+    ///   `output`
+    ///
+    /// Builder methods are provided to change these defaults and otherwise configure the process.
+    ///
+    /// ## Sftp subsystem
+    ///
+    /// To use sftp subsystem, it is recommended to enable feature `openssh-sftp-client`,
```
jonhoo, line 185: There is no such feature as far as I can tell?

jonhoo: Also, we should probably link to the `openssh-sftp-client` crate here.

tests/openssh.rs
```diff
@@ -822,3 +822,30 @@ async fn local_socket_forward() {
         assert!(output.status.success());
     }
 }
+
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn test_sftp_subsystem() {
```
jonhoo, line 828: Should we try to write a file to `/tmp` or something too just to check that the whole machinery actually works?

src/stdio.rs
jonhoo: Excellent, thank you!

And totally understandable that things may take longer when you're busy — I've been a prime example of that :)

jonhoo (changes requested):

src/changelog.rs
NobodyXu: Done!

tests/openssh.rs
NobodyXu: Done!

I added writing and reading to `/tmp/file` to test the integration with `openssh-sftp-client` is working!

src/session.rs
NobodyXu: Doc updated!

src/session.rs
```diff
@@ -182,9 +182,11 @@ impl Session {
     ///
     /// ## Sftp subsystem
     ///
-    /// To use sftp subsystem, it is recommended to enable feature `openssh-sftp-client`,
+    /// To use sftp subsystem, it is recommended to use crate [`openssh-sftp-client`],
```
jonhoo, line 185:
```suggestion
    /// To use the sftp subsystem, you'll want to use [`openssh-sftp-client`],
```

jonhoo (changes requested):

jonhoo (approved):

PR #67, opened 2022-04-25
Fix `no_route` on macos

### LINKED ISSUE #63: Bug/Test failure on MacBook
On Macbook Air 2020 with MacOS `12.3.1`, I got the following error when running `check.sh` and `run_ci_tests.sh`:

```
failures:

---- no_route stdout ----
Custom { kind: ConnectionAborted, error: "connect to host 255.255.255.255 port 22: Address family not supp
orted by protocol family" }
thread 'no_route' panicked at 'assertion failed: `(left == right)`
  left: `ConnectionAborted`,
 right: `Other`', tests/openssh.rs:676:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    no_route
```

The test in question is [`no_route`](https://github.com/openssh-rust/openssh/blob/729fe31bc184e5d352a752d79b4233bfbf872c17/tests/openssh.rs#L676).

The error is created using [`Error::interpret_ssh_error`](https://github.com/openssh-rust/openssh/blob/729fe31bc184e5d352a752d79b4233bfbf872c17/src/error.rs#L152).

NobodyXu:
Fixed #63 using "192.0.2.1" plus timeout of 0ns.

The only drawback is that now `no_route` takes much longer to run.

### DIFF
--- a/.github/workflows/os-check.yml
+++ b/.github/workflows/os-check.yml
@@ -29,6 +29,19 @@ jobs:
           profile: minimal
           toolchain: stable
       - uses: actions/checkout@v2
+      - name: Create Cargo.lock
+        uses: actions-rs/cargo@v1
+        with:
+          command: update
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}-os-check-v0
       - name: cargo test
         uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/os-check.yml
+++ b/.github/workflows/os-check.yml
@@ -32,5 +32,5 @@ jobs:
       - name: cargo check
         uses: actions-rs/cargo@v1
         with:
-          command: check
+          command: test
           args: --all-features --all-targets
--- a/.github/workflows/os-check.yml
+++ b/.github/workflows/os-check.yml
@@ -33,15 +33,6 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: update
-      - uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}-os-check-v0
       - name: cargo test
         uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/os-check.yml
+++ b/.github/workflows/os-check.yml
@@ -14,7 +14,7 @@ on:
       - 'run_ci_tests.sh'
       - 'start_sshd.sh'
       - 'stop_sshd.sh'
-name: cargo check
+name: os check
 jobs:
   os-check:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/67)

tests/openssh.rs
```diff
@@ -669,11 +671,15 @@ async fn cannot_resolve() {
 
 #[tokio::test]
 async fn no_route() {
-    for err in connects_err("255.255.255.255").await {
+    let mut builder = SessionBuilder::default();
+
+    builder.connect_timeout(Duration::from_nanos(0));
```
jonhoo, line 676: I _think_ a duration of 0s is actually "special" and means basically no timeout. I think you may have more luck setting this to something like 1.

.github/workflows/os-check.yml
```diff
@@ -32,5 +32,5 @@ jobs:
       - name: cargo check
         uses: actions-rs/cargo@v1
         with:
-          command: check
+          command: test
```
jonhoo, line 35: This doesn't seem right — this was intentionally `check` as the name of the job implies.

jonhoo (changes requested):

.github/workflows/os-check.yml
NobodyXu: Right, l will change the name of the job to match what it does then.

IMO it is necessary to run cargo-test on the github action so that it doesn't accidentally break on MacOS.

tests/openssh.rs
NobodyXu: Done!

.github/workflows/os-check.yml
NobodyXu: Done!

jonhoo (approved):
I'm happy with this without the caching change :+1:

.github/workflows/os-check.yml
```diff
@@ -29,8 +29,21 @@ jobs:
...
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}-os-check-v0
+      - name: cargo test
+        uses: actions-rs/cargo@v1
```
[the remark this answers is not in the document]
NobodyXu: Done!

.github/workflows/os-check.yml
NobodyXu: Actually I've been developing on MacBook Air 2020 for some time and it's been working quite good.
With Docker Desktop, I can run all the integration tests without any problem.

jonhoo: Right, but that doesn't work on GitHub actions

.github/workflows/os-check.yml
```diff
@@ -29,8 +29,12 @@ jobs:
           profile: minimal
           toolchain: stable
       - uses: actions/checkout@v2
-      - name: cargo check
+      - name: Create Cargo.lock
         uses: actions-rs/cargo@v1
         with:
-          command: check
+          command: update
```
jonhoo, line 35: I don't think this explicit update step is needed?

NobodyXu: Nope, I forgot to remove it .

It's part of the code required for caching.

NobodyXu: I will open another PR to fix it.

Looks like I need to slow down a bit to ensure quality of the PR/code.

NobodyXu: Done!
#74 shall fix this.

PR #71, opened 2022-05-02
Add dep-autoupgrade to create PR whenever there is a new major release of any dependency

NobodyXu:
This PR uses [NobodyXu/dependencies-autoupdate](https://github.com/NobodyXu/dependencies-autoupdate) which is a fork.

### DIFF
--- a/.github/workflows/dep-autoupgrade.yml
+++ b/.github/workflows/dep-autoupgrade.yml
@@ -1,0 +1,37 @@
+name: depdendency auto upgrade format
+
+env:
+  CARGO_TERM_COLOR: always
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '30 5,18 * * *'
+  workflow_dispatch:
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout the head commit of the branch
+      uses: actions/checkout@v3
+      with:
+        persist-credentials: false
+    - name: Install cargo-bininstall
+      run: |
+        mkdir -p  ~/.cargo/bin/
+        cd ~/.cargo/bin/ && curl -L $url | tar zxf - && chmod +x cargo-binstall
+      env:
+        url: https://github.com/ryankurte/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-gnu.tgz
+    - name: Install cargo-edit
+      run: |
+        cargo binstall --no-confirm cargo-edit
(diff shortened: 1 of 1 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/71)

jonhoo:
I'm not a huge fan of this — in general I prefer to combine many upgrades into one release _especially_ if they're API-visible changes that require a major bump on our side too (though we should have few of those). One alternative is to at least add a [dependency badge](https://deps.rs/repo/github/jonhoo/openssh-rs).

NobodyXu:
> I'm not a huge fan of this — in general I prefer to combine many upgrades into one release _especially_ if they're API-visible changes that require a major bump on our side too (though we should have few of those). One alternative is to at least add a [dependency badge](https://deps.rs/repo/github/jonhoo/openssh-rs).

No, the new workflow does not automatically release this crate.

What it does is that, if one of our dependencies releases a new major version, it would create a Pull Request on this repositories so that we don't have check for new major releases ourselves.

jonhoo:
Yep, I understand its behavior, but I think it'll still end up then being a trickle of smaller updates, and then it'll be up to us to batch them anyway. Essentially, I think it'll be mostly noise that we then have to filter out. That said, if you think it's valuable, I'm not completely opposed to trying it out.

Why are we using a fork of the tool though?

NobodyXu:
> Yep, I understand its behavior, but I think it'll still end up then being a trickle of smaller updates, and then it'll be up to us to batch them anyway. Essentially, I think it'll be mostly noise that we then have to filter out. That said, if you think it's valuable, I'm not completely opposed to trying it out.

It only creates PR on major releases of the dependencies, so IMHO there might not be much PR opened.

One of the motivation for this CI is that we don't have to constantly watch for new major releases in our dependencies.

> Why are we using a fork of the tool though?

Because the upstream implementation contains bugs and its owner is not actively maintaining it.

BTW, I added this CI to other openssh-rust
repositories and so far it works pretty good.

It hasn't created any PR as no major release has been released yet for their dependencies.

jonhoo:
Actually, how come we can't just use [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates)?

NobodyXu:
> Actually, how come we can't just use [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates)?

Thanks, I didn't know it before.

I will closed this PR and would open another PR that uses dependabot.

PR #73, opened 2022-05-08
Install `cargo-hack` on ci using `taiki/install-action`

NobodyXu:

### DIFF
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -24,12 +24,7 @@ jobs:
           profile: minimal
           toolchain: stable
       - uses: actions/checkout@v2
-      - name: Install cargo-hack
-        uses: actions-rs/install@v0.1
-        with:
-          crate: cargo-hack
-          version: latest
-          use-tool-cache: true
+      - uses: taiki-e/install-action@cargo-hack
       - name: cargo hack
         uses: actions-rs/cargo@v1
         with:


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/73)

jonhoo:
Looking at the [build logs](https://github.com/openssh-rust/openssh/actions/workflows/features.yml) it only seems to save 5-20s? We _should_ be getting most of the benefits from actions-rs/install already. I'm a little confused what value-add install-action provides, and why that change can't also be adopted into actions-rs/install?

NobodyXu:
> Looking at the [build logs](https://github.com/openssh-rust/openssh/actions/workflows/features.yml) it only seems to save 5-20s? We _should_ be getting most of the benefits from actions-rs/install already.

The "use-tool-cache" support in actions-rs/install is disabled, so it would just call cargo install to compile it every time the CI run.

> I'm a little confused what value-add install-action provides, and why that change can't also be adopted into actions-rs/install?

taiki/install-action guarantees that no compilation is done.

It simply copies the precompiled binary.

NobodyXu:
Another advantages of `taiki-e/install-action` always provides up-to-date `cargo-hack`, where using cache might end up using an out-dated version of `cargo-hack`.

jonhoo (approved):
Okay, fair enough. I still feel like this means that `actions-rs` should be fixed, but I'm also happy to use anything made by @taiki-e :)

PR #80, opened 2022-06-07
Feature/resume

### LINKED ISSUE #72: Let ssh decide its control sockets etc
Hi. I have a project where I'm trying to control ssh(&maybe sftp) from a Rust parent process, and I'd love to use your library.

However, the library seems very opinionated about passing `-S` etc options.

How would you feel about an API that lets `ssh` just do what it is configured to do, instead of forcing its hand so much? Right now, the library won't e.g. let me reuse a master connection that was opened from elsewhere.

NobodyXu:
Fixed #72 

This PR adds:
 - `Session::leak`
 - `Session::resume`
 - `Session::resume_mux`
 - `Session::force_terminate`

and integration tests for them.

This PR also added a minor change to `.github/workflows/style.yml` to make sure it failed on warnings like missing items and updated `run_ci_tests.sh` to remove temp dirs.

Edit:

It also [Fix random error in step "Test ssh connectivity"](https://github.com/openssh-rust/openssh/pull/80/commits/39298e7ce884be8175af6a74d58cef9bc81eae3a).

### DIFF
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -29,12 +29,13 @@ jobs:
           toolchain: stable
       - uses: actions/checkout@v2
       - run: |
+          # Wait for startup of openssh-server
+          sleep 15
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
           ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
         name: Test ssh connectivity
       - run: |
-          sleep 30
           eval $(ssh-agent)
           echo "SSH_AUTH_SOCK=$SSH_AUTH_SOCK" >> $GITHUB_ENV
           echo "SSH_AGENT_PID=$SSH_AGENT_PID" >> $GITHUB_ENV
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: |
           # Wait for startup of openssh-server
-          sleep 15
+          timeout 15 ./wait_for_sshd_start_up.sh
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
           ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -34,6 +34,7 @@ jobs:
           ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
         name: Test ssh connectivity
       - run: |
+          sleep 30
           eval $(ssh-agent)
           echo "SSH_AUTH_SOCK=$SSH_AUTH_SOCK" >> $GITHUB_ENV
           echo "SSH_AGENT_PID=$SSH_AGENT_PID" >> $GITHUB_ENV
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -55,4 +55,4 @@ jobs:
           command: doc
           args: --no-deps --all-features
         env:
-          RUSTDOCFLAGS: --cfg docsrs
+          RUSTDOCFLAGS: --cfg docsrs -Dwarnings
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -55,4 +55,4 @@ jobs:
           command: doc
           args: --no-deps --all-features
         env:
-          RUSTDOCFLAGS: --cfg docsrs -Dwarnings
+          RUSTDOCFLAGS: --cfg docsrs
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,8 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
       - uses: actions/checkout@v2
       - run: |
-          sleep 30
+          # Wait for startup of openssh-server
+          sleep 15
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
           ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: |
           # Wait for startup of openssh-server
+          timeout 15 ./wait_for_sshd_start_up.sh
           sleep 15
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,6 @@ jobs:
       - run: |
           # Wait for startup of openssh-server
           timeout 15 ./wait_for_sshd_start_up.sh
-          sleep 15
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
           ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,7 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
       - uses: actions/checkout@v2
       - run: |
+          sleep 30
           chmod 600 .test-key
           mkdir /tmp/openssh-rs
           ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
--- a/src/session.rs
+++ b/src/session.rs
@@ -317,8 +317,9 @@ impl Session {
     }
 
     /// Terminate the remote connection.
-    /// It would terminate the ssh multiplex server
-    /// regardless of how it is created.
+    ///
+    /// This destructor terminates the ssh multiplex server
+    /// regardless of how it was created.
     pub async fn close(self) -> Result<(), Error> {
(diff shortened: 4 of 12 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/80)

NobodyXu:
> # [Codecov](https://codecov.io/gh/openssh-rust/openssh/pull/80?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None) Report
> 
> > Merging [#80](https://codecov.io/gh/openssh-rust/openssh/pull/80?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None) ([bd85a0b](https://github.com/openssh-rust/openssh/commit/bd85a0bc6c62e0900a04e36aee1f4eec06cf98c0)) into [master](https://codecov.io/gh/openssh-rust/openssh/commit/13e80b7496bea9f40d27806a6b7ac84edfc3b867?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None) ([13e80b7](https://github.com/openssh-rust/openssh/commit/13e80b7496bea9f40d27806a6b7ac84edfc3b867)) will **increase** coverage by `0.58%`.
> > The diff coverage is `97.29%`.
> 
> [Impacted Files](https://codecov.io/gh/openssh-rust/openssh/pull/80?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None) 	Coverage Δ 	
> [src/native_mux_impl/session.rs](https://codecov.io/gh/openssh-rust/openssh/pull/80/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None#diff-c3JjL25hdGl2ZV9tdXhfaW1wbC9zZXNzaW9uLnJz) 	`93.75% <96.15%> (+0.52%)` 	⬆️
> [src/process_impl/session.rs](https://codecov.io/gh/openssh-rust/openssh/pull/80/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None#diff-c3JjL3Byb2Nlc3NfaW1wbC9zZXNzaW9uLnJz) 	`87.02% <97.22%> (+1.58%)` 	⬆️
> [src/session.rs](https://codecov.io/gh/openssh-rust/openssh/pull/80/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None#diff-c3JjL3Nlc3Npb24ucnM=) 	`74.28% <100.00%> (+5.32%)` 	⬆️

So happy to finally see the cov report in PR!

.github/workflows/style.yml
```diff
@@ -55,4 +55,4 @@ jobs:
           command: doc
           args: --no-deps --all-features
         env:
-          RUSTDOCFLAGS: --cfg docsrs
+          RUSTDOCFLAGS: --cfg docsrs -Dwarnings
```
jonhoo, line 58: I'm not a fan of denying warnings, even in CI because new warnings will _always_ accumulate over time simply as they get added to rustc/rustdoc/clippy. This is why we have GitHub annotations for PRs, which indicate the warnings without blocking the PR.

src/session.rs
```diff
@@ -65,6 +65,31 @@ impl From<native_mux_impl::Session> for Session {
...
+    ///
+    /// This connects to the ssh multiplex master using process mux impl.
+    #[cfg(feature = "process-mux")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "process-mux")))]
+    pub fn resume(ctl: Box<Path>, master_log: Option<Box<Path>>) -> Self {
+        process_impl::Session::resume(ctl, master_log).into()
+    }
+
+    /// Same as [`Session::resume`] except that it connects to
+    /// the ssh multiplex master using process mux impl.
```
jonhoo, line 86: I think this should say "native mux"?

src/session.rs
```diff
@@ -295,4 +320,16 @@ impl Session {
     pub async fn close(self) -> Result<(), Error> {
         delegate!(self.0, imp, { imp.close().await })
     }
+
+    /// Leak the underlying ssh multiplex master.
+    /// Return (path to control socket, path to ssh multiplex output log)
+    pub fn leak(self) -> (Box<Path>, Option<Box<Path>>) {
```
jonhoo, line 326: I think we should avoid calling this "leak", since that has a pretty strong association with leaking memory in Rust (see `Box::leak`). One option is to instead just have a configuration method for "what should happen on drop" — something like `close_on_drop(yes: bool)`, which the user can toggle on or off. If you want to keep it a destructing method like this, I'd go with `fn detach`.

src/session.rs
```diff
@@ -295,4 +320,16 @@ impl Session {
...
+        delegate!(self.0, imp, { imp.leak() })
+    }
+
+    /// Force terminate the remote connection, even if it is created
+    /// by [`Session::resume`] or [`Session::resume_mux`].
+    pub async fn force_terminate(self) -> Result<(), Error> {
```
jonhoo, line 332: Here, too, I think we can make the API a little clearer. One option is to rely on the same kind of configuration as mentioned above (`close_on_drop(yes: bool)`), where that defaults to no for resumed sessions, but can be set to yes. Alternatively, I think it's reasonable to make an explicit call to `close` _always_ mean `-O exit` (including for resumed sessions), and to _just_ change the default behavior of drop depending on whether the session was resumed or not.

jonhoo (changes requested):

.github/workflows/minimal.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           ssh -i .test-key -v -p 2222 -l test-user 127.0.0.1 -o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/tmp/openssh-rs/known_hosts whoami
         name: Test ssh connectivity
       - run: |
+          sleep 30
```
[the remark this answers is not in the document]
NobodyXu: I added this because the ci randomly failed.

The `openssh-server` container actually calls `apk` to install some softwares required to fetch the mod `DOCKER_MODS: linuxserver/mods:openssh-server-ssh-tunnel` and then downloads and applies the mod on the fly before starting `sshd` process, so the startup time varies and sometimes it just falls.

I tried `sleep 5` but it still failed so I changed that to `sleep 30`.

I guess I can try `sleep 15` as 15s should be enough.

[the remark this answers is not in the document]
NobodyXu: Done!

.github/workflows/style.yml
NobodyXu: Done!

src/session.rs
NobodyXu: Done!

src/session.rs
NobodyXu: I'd like to keep it as it return the path to control socket and the path to the master_log, which is useful for process mux implementation.

NobodyXu: Done!

src/session.rs
NobodyXu: Done!

tests/openssh.rs
```diff
@@ -829,6 +829,96 @@ async fn local_socket_forward() {
...
+        let (ctl2, master_log2) = session2.detach();
+
+        // Second resume_mux to ensure close handles tempdir set to None correctly
+        let session3 = Session::resume_mux(ctl2, master_log2);
+        session3.check().await.unwrap();
+
+        session3.close().await.unwrap();
+    }
+
+    // test force_terminate
```
jonhoo, line 902: This doesn't seem accurate any more since there's no `force_terminate` now.

src/session.rs
```diff
@@ -65,6 +65,31 @@ impl From<native_mux_impl::Session> for Session {
...
 impl Session {
+    /// Resume the connection using path to control socket and
+    /// path to ssh multiplex output log.
+    ///
+    /// If you do not use `-E` option (or redirection) to write
+    /// the log of the ssh multiplex master to the disk, you can
+    /// simply pass `None` to `master_log`.
+    ///
+    /// [`Session`] created this way will not be terminated on drop,
+    /// but can be forced terminated by [`Session::force_terminate`].
```
jonhoo, line 76: This should be called `close` now I think.

jonhoo: We should also update the documentation on `close` to mention that it will terminate the control master even if the session was resumed.

jonhoo (changes requested):
Nice, almost there! I like the simplicity of where we ended up. Just some cleanup related to old references to `force_terminate` left.

src/session.rs
NobodyXu: Done!

tests/openssh.rs
NobodyXu: Done!

.github/workflows/minimal.yml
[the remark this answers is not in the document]
Huanzo: To maybe get rid of a hard coded sleep one could use `ssh-keyscan` to probe the ssh server. I use `until ssh-keyscan "$IP" > /root/.ssh/known_hosts; do sleep 1; done` in one of my own CIs to wait for the remote host to accept ssh connections.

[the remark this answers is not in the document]
NobodyXu: Nice!
I will have a try.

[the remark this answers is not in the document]
NobodyXu: @Huanzo Thank you very much!
I've now fixed it by using a proper solution, combining your solution with a `timeout`.

src/session.rs
```diff
@@ -292,7 +317,17 @@ impl Session {
     }
 
     /// Terminate the remote connection.
+    /// It would terminate the ssh multiplex server
+    /// regardless of how it is created.
```
jonhoo, line 321:
```suggestion
    ///
    /// This destructor terminates the ssh multiplex server
    /// regardless of how it was created.
```

.github/workflows/test.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: |
           # Wait for startup of openssh-server
+          timeout 15 ./wait_for_sshd_start_up.sh
           sleep 15
```
jonhoo, line 34: Did you mean to remove the `sleep 15` here too?

jonhoo (changes requested):
Another two changes and I think we're there!

.github/workflows/test.yml
NobodyXu: Oops, good catch!

NobodyXu: Done!

jonhoo (approved):
Great, thank you!

jonhoo:
Released in 0.9.3 :tada:

PR #82, opened 2022-06-14
Rm `From<OwnedFd> for Stdio` impl

NobodyXu:
to avoid using `io_lifetimes` in public API.

### DIFF
--- a/src/stdio.rs
+++ b/src/stdio.rs
@@ -75,12 +75,6 @@ impl From<Stdio> for process::Stdio {
     }
 }
 
-impl From<OwnedFd> for Stdio {
-    fn from(fd: OwnedFd) -> Self {
-        Self(StdioImpl::Fd(fd))
-    }
-}
-
 macro_rules! impl_from_for_stdio {
     ($type:ty) => {
         impl From<$type> for Stdio {


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/82)

NobodyXu:
Looks like io-safety is finally going to be stablised in 1.63 https://github.com/rust-lang/rust/pull/95118

@jonhoo Should we wait for it instead and replace the crate io-lifetimes with the std counterparts?

src/stdio.rs
```diff
@@ -75,12 +75,6 @@ impl From<Stdio> for process::Stdio {
     }
 }
 
-impl From<OwnedFd> for Stdio {
```
jonhoo, line 78: Oh, nice, so we weren't even using it!

jonhoo (approved):
I'd like to remove it anyway in the interrim since as it stands it's a breaking change problem waiting to happen. Happy to add it back after 1.63 without the dependency.

NobodyXu:
> I'd like to remove it anyway in the interrim since as it stands it's a breaking change problem waiting to happen. Happy to add it back after 1.63 without the dependency.

@jonhoo  Shall we create a new release and yield all the other `0.9.*` release to ensure that no one depends on it?

jonhoo:
I don't think this is quite worth the (user) friction of a yank given that there isn't actually a security risk to this. The fact that there is a breaking change is going to be mostly irrelevant to users.

jonhoo:
Released as 0.9.2 :tada:

PR #85, opened 2022-06-30
Add documentation link in Cargo.toml

DJDuque:
Whenever you search for openssh in crates.io, this will add a "Documentation" link under the crate's name. This removes the annoying extra click of going into the crate's README if you know you only want to get to the documentation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ license = "MIT OR Apache-2.0"
 readme = "README.md"
 description = "SSH through OpenSSH"
 repository = "https://github.com/openssh-rust/openssh.git"
+documentation = "https://docs.rs/openssh"
 
 keywords = ["ssh","remote","openssh","orchestration"]
 categories = ["network-programming", "api-bindings"]


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/85)

NobodyXu (approved):

NobodyXu:
Wait, why is it closed?
Did I accidentally close it?

jonhoo:
Ah, so, this is actually a shortcoming of crates.io: https://github.com/rust-lang/crates.io/issues/1484. I'm hesitant to add a `documentation` link like that because it'll always point to the _latest_ documentation, which will be wrong if someone is looking at an older version of `openssh` on crates.io.

DJDuque:
That is a fair point, although I believe that the latest documentation is still a reasonable entry point; it reduces the number of clicks needed for everybody. People looking for older versions still need to go `README -> Versions -> Select older version -> Documentation link of that older version`; here they could just go `Latest documentation -> from the top left, choose whatever version`.

But leaving out the documentation link is also reasonable.

jonhoo (approved):
Fair enough, I buy that. Let's merge it!

jonhoo:
Released as 0.9.4 :tada:

PR #86, opened 2022-07-12
Optimize and make `request_port_forward` easier to use

NobodyXu:
- Avoid dup monomorphization.
 - Rm field `addr` in `process_impl::Session`
 - Take `impl Into<...>` in `Session::request_port_forwarding`
 - Implement `From<...> for Socket`
 -  Optimize Session::close: Avoid dup code
    - Avoid duplicate call to `close_impl` in both `process_impl` and
      `native_mux_impl`
    - Extract out code to close `TempDir` out into `Session::close`

### DIFF
--- a/src/native_mux_impl/session.rs
+++ b/src/native_mux_impl/session.rs
@@ -51,17 +51,13 @@ impl Session {
 
     pub(crate) async fn request_port_forward(
         &self,
-        forward_type: impl Into<ForwardType>,
-        listen_socket: impl Into<Socket<'_>>,
-        connect_socket: impl Into<Socket<'_>>,
+        forward_type: ForwardType,
+        listen_socket: Socket<'_>,
+        connect_socket: Socket<'_>,
     ) -> Result<(), Error> {
         Connection::connect(&self.ctl)
             .await?
-            .request_port_forward(
-                forward_type.into(),
-                &listen_socket.into(),
-                &connect_socket.into(),
-            )
+            .request_port_forward(forward_type, &listen_socket, &connect_socket)
             .await?;
 
         Ok(())
--- a/src/native_mux_impl/session.rs
+++ b/src/native_mux_impl/session.rs
@@ -76,17 +76,13 @@ impl Session {
         Ok(())
     }
 
-    pub(crate) async fn close(mut self) -> Result<(), Error> {
-        // This also set self.tempdir to None so that Drop::drop would do nothing.
-        if let Some(tempdir) = self.tempdir.take() {
-            self.close_impl().await?;
+    pub(crate) async fn close(mut self) -> Result<Option<TempDir>, Error> {
+        // Take self.tempdir so that drop would do nothing
+        let tempdir = self.tempdir.take();
 
-            tempdir.close().map_err(Error::Cleanup)?;
-        } else {
-            self.close_impl().await?;
-        }
+        self.close_impl().await?;
 
-        Ok(())
+        Ok(tempdir)
     }
 
     pub(crate) fn detach(mut self) -> (Box<Path>, Option<Box<Path>>) {
--- a/src/native_mux_impl/session.rs
+++ b/src/native_mux_impl/session.rs
@@ -1,4 +1,4 @@
-use super::{Command, Error, ForwardType, Socket};
+use super::{Command, Error};
 
 use std::ffi::OsStr;
 use std::os::unix::ffi::OsStrExt;
@@ -51,13 +51,17 @@ impl Session {
 
     pub(crate) async fn request_port_forward(
         &self,
-        forward_type: ForwardType,
-        listen_socket: Socket<'_>,
-        connect_socket: Socket<'_>,
+        forward_type: crate::ForwardType,
+        listen_socket: crate::Socket<'_>,
+        connect_socket: crate::Socket<'_>,
     ) -> Result<(), Error> {
         Connection::connect(&self.ctl)
             .await?
-            .request_port_forward(forward_type, &listen_socket, &connect_socket)
+            .request_port_forward(
+                forward_type.into(),
+                &listen_socket.into(),
+                &connect_socket.into(),
+            )
             .await?;
 
         Ok(())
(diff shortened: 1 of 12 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/86)

NobodyXu:
Looks like the CI failed mostly because it cannot add the host to `known_hosts`.

Since both `run_ci_tests.sh` and `check.sh` succeeded on my MacBook, I assumed this to be some problem with the CI setup.

I will probably fix this tomorrow.

NobodyXu:
Looking at the CI log, the failures were mostly caused by `session.close()`, which means that `ssh` was able to run just fine but it cannot add the addr to `known_hosts`.

NobodyXu:
This error does not occur on my MacBook, which uses openssh 9.0
The CI uses openssh 8.2, which is probably why I cannot reproduce it locally.

NobodyXu:
> Hmmm, I thought this was because of removal of field `addr` in `process_impl::Session`, but it fails none the less. I think this might be CI setup problem.

I tried to reproduce on my own ubuntu by manually installing `OpenSSH_8.2p1` from source and all test cases passed without problem, so that is definitely a CI setup problem.

NobodyXu:
I think the key problem is this:

```
thread 'bad_remote_command' panicked at 'called `Result::unwrap()` on an `Err` value: Master(Custom { kind: Other, error: "Failed to add the host to the list of known hosts (/run/user/1001/openssh-rs/known_hosts)." })', tests/openssh.rs:425:31
```

It is trying to add the host to somewhere under `/run/user/1001/openssh-rs` which is a non-existent directory.
IDK why that worked just fine before but suddenly failed now, even with the same openssh version.

My guess is that the CI now set `XDG_RUNTIME_DIR`.

NobodyXu:
Finally!
The CI is fixed.

jonhoo (changes requested):
Nice, I like this change!

I don't think the monomorphization changes are worth it though. They make the code more complex without (I suspect) improving build times much. In general, the inner function trick only makes a difference if the body of the inner function gets decently large/complicated, but that isn't the case for any of these.

jonhoo:
> I do think we need to avoid duplicate monomorphization of the `{process_impl, native_mux_impl}::{Session, Command}`, since they provides the actual implementation and can be quite complex.

I'm not sure I follow — the inner trick only removes monomorphization of the body of _that_ inner function. Functions that the inner function then calls as handled separately. So turning
```rust
fn generic<T: Into<Foo>>(t: T) {
    complex(t.into())
}
```
into
```rust
fn generic<T: Into<Foo>>(t: T) {
    fn inner(f: Foo) { complex(f) }
    inner(t.into())
}
```
will basically have no effect since the _only_ change is how the _call_ to `complex` gets compiled not `complex` itself.

You _do_ want to make sure that if you have
```rust
fn complex<t: impl Into<Foo>>(t) {}
```
you call it as
```rust
fn generic<T: Into<Foo>>(t: T) {
    complex(t.into())
```
not
```rust
fn generic<T: Into<Foo>>(t: T) {
    complex(t)
```
as otherwise `complex` _will_ be compiled once for each `T`, but that doesn't require the inner function trick.

jonhoo:
Could you point me to which of them you think should be kept? All the inner functions in this PR look very small to me.

NobodyXu:
> Could you point me to which of them you think should be kept? All the inner functions in this PR look very small to me.

`SessionBuilder::{connect, connect_mux}` and `Session::{connect, connet_mux}`.

I will probably revert the change to later as well.

src/port_forwarding.rs
```diff
@@ -46,16 +46,69 @@ pub enum Socket<'a> {
...
-        let addr = it.next().ok_or_else(|| {
-            io::Error::new(io::ErrorKind::Other, "no more socket addresses to try")
-        })?;
+    pub fn new<T: ToSocketAddrs>(addr: &T) -> Result<Socket<'static>, io::Error> {
+        fn inner(it: &mut dyn Iterator<Item = SocketAddr>) -> Result<Socket<'static>, io::Error> {
+            it.next()
+                .ok_or_else(|| {
+                    io::Error::new(io::ErrorKind::Other, "no more socket addresses to try")
+                })
+                .map(Socket::TcpSocket)
```
jonhoo, line 108: This function body is still pretty trivial to compile, so I don't think the inner function trick is worthwhile here.

src/process_impl/session.rs
```diff
@@ -10,23 +10,24 @@ use tokio::process;
 
 use tempfile::TempDir;
 
+/// Session does not have field "addr" because ssh don't care about
+/// the addr as long as we have the ctl.
+/// It is tested on OpenSSH 8.2p1, 8.9p1, 9.0p1
```
jonhoo, line 15: I don't think we need this comment really.

src/native_mux_impl/session.rs
```diff
@@ -76,17 +76,13 @@ impl Session {
...
-            self.close_impl().await?;
+    pub(crate) async fn close(mut self) -> Result<Option<TempDir>, Error> {
+        // Take self.tempdir so that drop would do nothing
+        let tempdir = self.tempdir.take();
 
-            tempdir.close().map_err(Error::Cleanup)?;
-        } else {
-            self.close_impl().await?;
-        }
+        self.close_impl().await?;
```
jonhoo, line 83: Doesn't this change the semantics of `close`? Now we _always_ call` close_impl`, whereas previously we would only call it if `tempdir` was already `None`?

src/port_forwarding.rs
NobodyXu: Good catch!

I missed that one.

src/native_mux_impl/session.rs
NobodyXu: Actually, `close` would always call `close_impl`.

src/port_forwarding.rs
NobodyXu: Done!

src/process_impl/session.rs
NobodyXu: This comment is removed and I added that to `new_std_cmd` where it is actually relevant.

jonhoo (approved):

jonhoo:
Released in 0.9.5 :tada:

PR #94, opened 2022-08-13
Use Rust 1.63's `OwnedFd`

NobodyXu:
- Bump msrv to 1.63
 - Remove dep `io_lifetimes` and `libc`
 - Add back `impl From<std::os::unix::io::OwnedFd> for Stdio`


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/94)

jonhoo (approved):
The MSRV bump is unfortunate, but I think it's totally worth it in this case.

PR #104, opened 2022-10-29
Fixed stdout/stderr cut off bug when using `RemoteChild::wait_with_output`

### LINKED ISSUE #103: `wait_with_output` + `native-mux` cuts off stdout output
I have code that's roughly like follows:

```rust
// …
let std::process::Output {
        status,
        stderr,
        stdout,
    } = c.spawn().await?.wait_with_output().await?;
// …
let content = &stdout;
tracing::warn!(
            "daily content is {} long and ends with {:?}",
            content.len(),
            content.iter().rev().next()
        );
// …
```

The command being ran on the remote produces relatively large amount of output on a single line. When I'm using `native-mux` feature, this output is incomplete!

With `native-mux` I see:

```
2022-10-25T05:43:55.091989Z  WARN start_sim::live: daily content is 147456 long and ends with Some(110)
```

but if I change to `process-mux` and change `connect_mux` to `connect`, I see

```
2022-10-25T05:42:40.638584Z  WARN start_sim::live: daily content is 410777 long and ends with Some(10)
```

The `process-mux` is correct.

I don't have minimum reproducible example at hand but I suspect one can just create a large file (say 1MiB) on a remote and spawn `cat large-file` via `openssh` and reproduce the issue.

I tried with 0.9.7, same result.

NobodyXu:
Fixed #103 

This PR fixed two bugs:
 - Fix `Fd::as_raw_fd_or_null_fd` in native-mux: Disable non-blocking on all fds before passing them to the ssh multiplex server
 - Fix `RemoteChild::wait_with_output`: Await on `self.wait()` after stdout/stderr is fully read in, otherwise the ssh multiplex server would stop writing data to stdout/stderr once `self.wait()` is done (which drops the connection to the server)

### DIFF
--- a/src/child.rs
+++ b/src/child.rs
@@ -175,8 +175,8 @@ impl<'s> RemoteChild<'s> {
         let (stdout, stderr) = try_join!(stdout_read, stderr_read)?;
         Ok(Output {
             // Once self.wait().await is done, the connection to the multiplex
-            // server would also be dropped and the server would stop
-            // sending data to stdout/stderr.
+            // server will also be dropped and the server will stop sending
+            // data to stdout/stderr.
             status: self.wait().await?,
             stdout,
             stderr,
--- a/src/child.rs
+++ b/src/child.rs
@@ -63,7 +63,7 @@ macro_rules! delegate {
 /// [`Command`](crate::Command).
 ///
 /// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be sent to the
-/// remote process and `stdout` and `stderr` would eof immediately.
+/// remote process and `stdout` and `stderr` will yield EOF immediately.
 ///
 /// ```rust,no_run
 /// # async fn foo() {
--- a/src/child.rs
+++ b/src/child.rs
@@ -62,7 +62,7 @@ macro_rules! delegate {
 /// available,`Stdio::piped()` should be passed to the corresponding method on
 /// [`Command`](crate::Command).
 ///
-/// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be send to the
+/// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be sent to the
 /// remote process and `stdout` and `stderr` would eof immediately.
 ///
 /// ```rust,no_run
--- a/src/child.rs
+++ b/src/child.rs
@@ -62,6 +62,9 @@ macro_rules! delegate {
 /// available,`Stdio::piped()` should be passed to the corresponding method on
 /// [`Command`](crate::Command).
 ///
+/// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be send to the
+/// remote process and `stdout` and `stderr` would eof immediately.
+///
 /// ```rust,no_run
 /// # async fn foo() {
 /// # let child: openssh::RemoteChild<'static> = unimplemented!();
--- a/src/child.rs
+++ b/src/child.rs
@@ -174,9 +174,12 @@ impl<'s> RemoteChild<'s> {
         // and cause the remote process to block forever.
         let (stdout, stderr) = try_join!(stdout_read, stderr_read)?;
         Ok(Output {
-            // Once self.wait().await is done, the connection to the multiplex
-            // server will also be dropped and the server will stop sending
-            // data to stdout/stderr.
+            // The self.wait() future terminates the stdout and stderr futures
+            // when it resolves, even if there may still be more data arriving
+            // from the server.
+            //
+            // Therefore, we wait for them first, and only once they're complete
+            // do we wait for the process to have terminated.
             status: self.wait().await?,
             stdout,
             stderr,
--- a/src/child.rs
+++ b/src/child.rs
@@ -169,9 +169,12 @@ impl<'s> RemoteChild<'s> {
 
         // Execute them concurrently to avoid the pipe buffer being filled up
         // and cause the remote process to block forever.
-        let (status, stdout, stderr) = try_join!(self.wait(), stdout_read, stderr_read)?;
+        let (stdout, stderr) = try_join!(stdout_read, stderr_read)?;
         Ok(Output {
-            status,
+            // Once self.wait().await is done, the connection to the multiplex
+            // server would also be dropped and the server would stop
+            // sending data to stdout/stderr.
+            status: self.wait().await?,
             stdout,
             stderr,
         })
--- a/src/native_mux_impl/command.rs
+++ b/src/native_mux_impl/command.rs
@@ -81,6 +81,15 @@ impl Command {
             .into_iter()
             .zip(is_inherited)
             .filter_map(|(stdio, is_inherited)| if !is_inherited { Some(stdio) } else { None })
+            // Note that once we do this, these file descriptors
+            // (and the Fds we got them from above) should no longer be used in
+            // any async context, as they'd start blocking.
+            //
+            // We give away the descriptors in stdios when we pass them to
+            // open_new_session (which doesn't use them in an async context),
+            // and the Fds are dropped when this function returns, meaning no
+            // owned file descriptors we set to be blocking here can be used in
+            // an async context in the future.
             .try_for_each(|stdio| set_blocking(stdio).map_err(Error::ChildIo))?;
 
         let cmd = NonZeroByteSlice::new(&self.cmd).ok_or(Error::InvalidCommand)?;
--- a/src/native_mux_impl/command.rs
+++ b/src/native_mux_impl/command.rs
@@ -1,6 +1,6 @@
 use super::Error;
 use super::RemoteChild;
-use super::{ChildStderr, ChildStdin, ChildStdout, Stdio};
+use super::{stdio::set_blocking, ChildStderr, ChildStdin, ChildStdout, Stdio};
 
 use std::borrow::Cow;
 use std::ffi::OsStr;
@@ -70,6 +70,10 @@ impl Command {
             stdout.as_raw_fd_or_null_fd()?,
             stderr.as_raw_fd_or_null_fd()?,
         ];
+
+        for stdio in stdios {
+            set_blocking(stdio).map_err(Error::ChildIo)?;
+        }
 
         let cmd = NonZeroByteSlice::new(&self.cmd).ok_or(Error::InvalidCommand)?;
 
--- a/src/native_mux_impl/command.rs
+++ b/src/native_mux_impl/command.rs
@@ -1,6 +1,6 @@
 use super::Error;
 use super::RemoteChild;
-use super::{stdio::set_blocking, ChildStderr, ChildStdin, ChildStdout, Stdio};
+use super::{ChildStderr, ChildStdin, ChildStdout, Stdio};
 
 use std::borrow::Cow;
 use std::ffi::OsStr;
@@ -70,27 +70,6 @@ impl Command {
             stdout.as_raw_fd_or_null_fd()?,
             stderr.as_raw_fd_or_null_fd()?,
         ];
-
-        let is_inherited = [
-            self.stdin_v.is_inherited(),
-            self.stdout_v.is_inherited(),
-            self.stderr_v.is_inherited(),
-        ];
-
-        stdios
-            .into_iter()
-            .zip(is_inherited)
-            .filter_map(|(stdio, is_inherited)| if !is_inherited { Some(stdio) } else { None })
-            // Note that once we do this, these file descriptors
-            // (and the Fds we got them from above) should no longer be used in
-            // any async context, as they'd start blocking.
-            //
-            // We give away the descriptors in stdios when we pass them to
-            // open_new_session (which doesn't use them in an async context),
-            // and the Fds are dropped when this function returns, meaning no
-            // owned file descriptors we set to be blocking here can be used in
-            // an async context in the future.
-            .try_for_each(|stdio| set_blocking(stdio).map_err(Error::ChildIo))?;
 
         let cmd = NonZeroByteSlice::new(&self.cmd).ok_or(Error::InvalidCommand)?;
 
--- a/src/native_mux_impl/command.rs
+++ b/src/native_mux_impl/command.rs
@@ -71,9 +71,17 @@ impl Command {
             stderr.as_raw_fd_or_null_fd()?,
         ];
 
-        for stdio in stdios {
-            set_blocking(stdio).map_err(Error::ChildIo)?;
-        }
+        let is_inherited = [
+            self.stdin_v.is_inherited(),
+            self.stdout_v.is_inherited(),
+            self.stderr_v.is_inherited(),
+        ];
+
+        stdios
+            .into_iter()
+            .zip(is_inherited)
+            .filter_map(|(stdio, is_inherited)| if !is_inherited { Some(stdio) } else { None })
+            .try_for_each(|stdio| set_blocking(stdio).map_err(Error::ChildIo))?;
 
         let cmd = NonZeroByteSlice::new(&self.cmd).ok_or(Error::InvalidCommand)?;
 
--- a/src/native_mux_impl/stdio.rs
+++ b/src/native_mux_impl/stdio.rs
@@ -68,6 +68,10 @@ impl Fd {
         }
     }
 
+    /// # Safety
+    ///
+    /// `T::into_raw_fd` must return a valid fd and transfers
+    /// the ownershipt of it.
     unsafe fn new_owned<T: IntoRawFd>(fd: T) -> Self {
         let raw_fd = fd.into_raw_fd();
         Fd::Owned(OwnedFd::from_raw_fd(raw_fd))
--- a/src/native_mux_impl/stdio.rs
+++ b/src/native_mux_impl/stdio.rs
@@ -42,11 +42,15 @@ fn cvt(ret: c_int) -> io::Result<c_int> {
     }
 }
 
-fn set_blocking(fd: RawFd) -> io::Result<()> {
+fn set_blocking_inner(fd: RawFd) -> io::Result<()> {
     let flags = cvt(unsafe { fcntl(fd, F_GETFL) })?;
     cvt(unsafe { fcntl(fd, F_SETFL, flags & (!O_NONBLOCK)) })?;
 
     Ok(())
+}
+
+fn set_blocking(fd: RawFd) -> Result<(), Error> {
+    set_blocking_inner(fd).map_err(Error::ChildIo)
 }
 
 impl Fd {
@@ -66,12 +70,7 @@ impl Fd {
     /// the ownershipt of it.
     unsafe fn new_owned<T: IntoRawFd>(fd: T) -> Result<Self, Error> {
         let raw_fd = fd.into_raw_fd();
-        // Create owned_fd so that the fd will be closed on error
-        let owned_fd = OwnedFd::from_raw_fd(raw_fd);
-
-        set_blocking(raw_fd).map_err(Error::ChildIo)?;
-
-        Ok(Fd::Owned(owned_fd))
+        Ok(Fd::Owned(OwnedFd::from_raw_fd(raw_fd)))
     }
 }
 
@@ -106,12 +105,16 @@ impl Stdio {
             StdioImpl::Null => Ok((Fd::Null, None)),
             StdioImpl::Pipe => {
                 let (read, write) = create_pipe()?;
+
+                // read end will be sent to ssh multiplex server
+                // and it expects blocking fd.
+                set_blocking(read.as_raw_fd())?;
                 Ok((read.try_into()?, Some(write)))
             }
             StdioImpl::Fd(fd, owned) => {
                 let raw_fd = fd.as_raw_fd();
                 if *owned {
-                    set_blocking(raw_fd).map_err(Error::ChildIo)?;
+                    set_blocking(raw_fd)?;
                 }
                 Ok((Fd::Borrowed(raw_fd), None))
             }
@@ -124,12 +127,16 @@ impl Stdio {
             StdioImpl::Null => Ok((Fd::Null, None)),
             StdioImpl::Pipe => {
                 let (read, write) = create_pipe()?;
+
+                // write end will be sent to ssh multiplex server
+                // and it expects blocking fd.
+                set_blocking(write.as_raw_fd())?;
                 Ok((write.try_into()?, Some(read)))
             }
             StdioImpl::Fd(fd, owned) => {
                 let raw_fd = fd.as_raw_fd();
                 if *owned {
-                    set_blocking(raw_fd).map_err(Error::ChildIo)?;
+                    set_blocking(raw_fd)?;
                 }
                 Ok((Fd::Borrowed(raw_fd), None))
             }
--- a/src/native_mux_impl/stdio.rs
+++ b/src/native_mux_impl/stdio.rs
@@ -42,7 +42,7 @@ fn cvt(ret: c_int) -> io::Result<c_int> {
     }
 }
 
-pub(super) fn set_blocking(fd: RawFd) -> io::Result<()> {
+fn set_blocking(fd: RawFd) -> io::Result<()> {
     let flags = cvt(unsafe { fcntl(fd, F_GETFL) })?;
     cvt(unsafe { fcntl(fd, F_SETFL, flags & (!O_NONBLOCK)) })?;
 
@@ -64,14 +64,21 @@ impl Fd {
     ///
     /// `T::into_raw_fd` must return a valid fd and transfers
     /// the ownershipt of it.
-    unsafe fn new_owned<T: IntoRawFd>(fd: T) -> Self {
+    unsafe fn new_owned<T: IntoRawFd>(fd: T) -> Result<Self, Error> {
         let raw_fd = fd.into_raw_fd();
-        Fd::Owned(OwnedFd::from_raw_fd(raw_fd))
+        // Create owned_fd so that the fd will be closed on error
+        let owned_fd = OwnedFd::from_raw_fd(raw_fd);
+
+        set_blocking(raw_fd).map_err(Error::ChildIo)?;
+
+        Ok(Fd::Owned(owned_fd))
     }
 }
 
-impl From<PipeRead> for Fd {
-    fn from(pipe_read: PipeRead) -> Self {
+impl TryFrom<PipeRead> for Fd {
+    type Error = Error;
+
+    fn try_from(pipe_read: PipeRead) -> Result<Self, Error> {
         // Safety:
         //
         // PipeRead::into_raw_fd returns a valid fd and transfers the
@@ -80,8 +87,10 @@ impl From<PipeRead> for Fd {
     }
 }
 
-impl From<PipeWrite> for Fd {
-    fn from(pipe_write: PipeWrite) -> Self {
+impl TryFrom<PipeWrite> for Fd {
+    type Error = Error;
+
+    fn try_from(pipe_write: PipeWrite) -> Result<Self, Error> {
         // Safety:
         //
         // PipeWrite::into_raw_fd returns a valid fd and transfers the
@@ -97,9 +106,15 @@ impl Stdio {
             StdioImpl::Null => Ok((Fd::Null, None)),
             StdioImpl::Pipe => {
                 let (read, write) = create_pipe()?;
-                Ok((read.into(), Some(write)))
+                Ok((read.try_into()?, Some(write)))
             }
-            StdioImpl::Fd(fd) => Ok((Fd::Borrowed(fd.as_raw_fd()), None)),
+            StdioImpl::Fd(fd, owned) => {
+                let raw_fd = fd.as_raw_fd();
+                if *owned {
+                    set_blocking(raw_fd).map_err(Error::ChildIo)?;
+                }
+                Ok((Fd::Borrowed(raw_fd), None))
+            }
         }
     }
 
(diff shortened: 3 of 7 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/104)

NobodyXu:
@jonhoo P.S. I would like to squash merge this PR and edit the commit messages since I inverted some commits.

NobodyXu:
@Fuuzetsu Can you test this PR to see if it fixed the bug for you please?

src/child.rs
```diff
@@ -62,6 +62,9 @@ macro_rules! delegate {
 /// available,`Stdio::piped()` should be passed to the corresponding method on
 /// [`Command`](crate::Command).
 ///
+/// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be send to the
```
jonhoo, line 65:
```suggestion
/// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be sent to the
```

src/child.rs
```diff
@@ -169,9 +172,12 @@ impl<'s> RemoteChild<'s> {
 
         // Execute them concurrently to avoid the pipe buffer being filled up
         // and cause the remote process to block forever.
-        let (status, stdout, stderr) = try_join!(self.wait(), stdout_read, stderr_read)?;
+        let (stdout, stderr) = try_join!(stdout_read, stderr_read)?;
         Ok(Output {
-            status,
+            // Once self.wait().await is done, the connection to the multiplex
+            // server would also be dropped and the server would stop
+            // sending data to stdout/stderr.
```
jonhoo, line 179:
```suggestion
            // Once self.wait().await is done, the connection to the multiplex
            // server will also be dropped and the server will stop sending
            // data to stdout/stderr.
```

jonhoo: I'm not sure I follow this comment. We only get to this line if both `stdout` _and_ `stderr` have already been closed (after all, we just joined them). In fact, I'm not sure why this change actually makes a difference, since the join shouldn't return until all three futures resolve. I'm guessing maybe waiting on the client has the side-effect of prematurely closing the stdout/stderr? If so, we should comment that here.

src/native_mux_impl/stdio.rs
```diff
@@ -28,25 +29,50 @@ fn get_null_fd() -> Result<RawFd, Error> {
...
+        let fd = match self {
+            Owned(owned_fd) => Some(owned_fd.as_raw_fd()),
+            Borrowed(rawfd) => Some(*rawfd),
+            Null => None,
+        };
 
-            Borrowed(rawfd) => Ok(*rawfd),
-            Null => get_null_fd(),
+        if let Some(fd) = fd {
+            set_blocking(fd).map_err(Error::ChildIo)?;
```
jonhoo, line 63: I think this is only okay to do for file descriptors we know we're going to give away. It'd be bad to do this for a file descriptor that we also continued using under `tokio`. Since this only takes `&self`, I don't actually know that we _know_ the return value will _only_ be given away after this function is called. Another alternative would be to `dup` the FD, set blocking on the dup'd FD, and then give that away.

src/native_mux_impl/stdio.rs
```diff
@@ -1,12 +1,13 @@
-use crate::stdio::StdioImpl;
-use crate::Error;
-use crate::Stdio;
+use crate::{stdio::StdioImpl, Error, Stdio};
 
-use once_cell::sync::OnceCell;
+use std::{
+    fs::{File, OpenOptions},
+    io,
+    os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, OwnedFd, RawFd},
```
jonhoo, line 6: This bumps MSRV by a decent amount, which is a little unfortunate, but not the end of the world.

src/native_mux_impl/stdio.rs
```diff
@@ -28,25 +29,50 @@ fn get_null_fd() -> Result<RawFd, Error> {
...
+
+            Ok(fd)
+        } else {
+            get_null_fd()
         }
     }
+
+    fn new_owned<T: IntoRawFd>(fd: T) -> Self {
+        let raw_fd = fd.into_raw_fd();
+        // Safety: IntoRawFd::into_raw_fd must return a valid raw fd.
```
jonhoo, line 73: That's not true. [The docs](https://doc.rust-lang.org/std/os/unix/io/trait.IntoRawFd.html#tymethod.into_raw_fd) explicitly say it's not a requirement of `IntoRawFd` and recommend another way to do it.

jonhoo (changes requested):

src/child.rs
NobodyXu: I observe that ssh multiplex server sends the process status to openssh-rs once it received the corresponding packet from sshd.

That packet could come before all stdout/stderr data is flushed, so self.wait().await could finish before all stdout/stderr data is completely read in, closing the connection to the ssh multiplex server and cause it to drop the channel and stop flushing data.

src/native_mux_impl/stdio.rs
NobodyXu: `Command::{stdin, stdout, stderr}` all takes `Stdio` by value.
And `Stdio` can be either `null`, `pipe` or constructed from a raw fd, which is considered taken by value, so it should be fine.

> Another alternative would be to dup the FD, set blocking on the dup'd FD, and then give that away.

`O_NONBLOCK` is shared by all fd pointing to the same file, so duplicating it has no use.

src/child.rs
jonhoo: Ohh, this is because with `native-mux` this `wait` isn't a process wait (which doesn't _signal_ a wait), it actively starts tear-down of the channel? That seems counter-intuitive if so. In general, the expectation is that `.wait` will _just_ wait, and not _also_ trigger shutdown.

src/native_mux_impl/stdio.rs
NobodyXu: We are already using `OwnedFd` in `impl From<OwnedFd> for Stdio`, so IMO it's fine.

src/child.rs
jonhoo: Ah, wait, I see what you're saying. It does feel very strange that the openssh server will send a close status message if it doesn't _also_ know that the remote process has terminated (which also means it has necessarily flushed its output). This all still implies to me that `wait` has a side-effect beyond just waiting for termination.

src/native_mux_impl/stdio.rs
NobodyXu: Unfortunately it doesn't work since `tokio_pipe::{PipeRead, PipeWrite}` does not implement `Into<OwnedFd>`

src/native_mux_impl/stdio.rs
jonhoo: Hmm, but doesn't that imply that we should take `self` here, not `&self`? To ensure we don't accidentally continue using the fd after giving away ownership in this way?

src/child.rs
NobodyXu: > This all still implies to me that wait has a side-effect beyond just waiting for termination.

Yes, that's one of the error I didn't forsee.

When I designed the api of openssh-mux-client, I didn't realize that the process status can come before all data is flushed or closing the connection to ssh multiplex server would make it drop the channel and any data that is yet to be flushed.

src/native_mux_impl/stdio.rs
jonhoo: In that case we should not take a generic `T` here, but instead have constructors _specifically_ for `PIpeRead` and `PipeWrite` where we know that _their_ implementations of `IntoRawFd` _do_ guarantee that ownership is transferred.

src/child.rs
NobodyXu: But that is going to be fixed in the next-gen proxy-client, of which the ssh multiplex server would act as a proxy and all the requests/responses are going to be produced and handled in openssh-rs.

That would make multiplexing multiple `RemoteChild`s on the same connection possible, and also providing new apis like killing remote process, receiving signal used to kill the remote process, etc.

NobodyXu: > 

Ohh, this is because with native-mux this wait isn't a process wait (which doesn't signal a wait), it actively starts tear-down of the channel? That seems counter-intuitive if so. In general, the expectation is that .wait will just wait, and not also trigger shutdown.

Yes and that is the problem.

I can change the api of mux-client, but since `RemoteChild::wait` takes `self` by value, I guess that won't help anyway.

src/native_mux_impl/stdio.rs
NobodyXu: Here `Fd` holds the `OwnedFd`, which is the write/read end of the pipe that needs to be held until the `RawFd` is sent to the ssh multiplex server.

src/native_mux_impl/stdio.rs
NobodyXu: Done!

To avoid duplicate code, I make `Fd::new_owned` an unsafe function and then implements `impl From<PipeRead> for Fd` and `impl From<PipeWrite> for Fd`.

src/child.rs
jonhoo: I think it's counter-intuitive for `RemoteChild::wait` to _also_ trigger termination. That's not generally how `wait` works, and probably now how people expect it to work. But you are right that it's likely a breaking change. The fact that it takes `self` shouldn't make a difference though — if it blocks until the status is received, it won't drop `self` (which I assume is what triggers the termination).

In either case, I'd like this comment to reflect what's going on a bit better. Maybe (if it's correct):

>  The `self.wait()` future terminates the stdout and stderr futures when it resolves, even if there may still be more data arriving from the server. Therefore, we wait for them _first_, and only once they're complete do we wait for the process to have terminated.

src/native_mux_impl/stdio.rs
jonhoo: But isn't this the method that produces the fd that we send to the server? If so, it should be fine for it to take `self`. If it isn't, then that implies that the `fd` is still going to be used going forward, in which case setting !non-blocking here may have unintended consequences.

jonhoo: Also, best to avoid marking these discussions as resolved until they actually are :)

src/native_mux_impl/stdio.rs
```diff
@@ -68,10 +68,29 @@ impl Fd {
         }
     }
 
-    fn new_owned<T: IntoRawFd>(fd: T) -> Self {
+    unsafe fn new_owned<T: IntoRawFd>(fd: T) -> Self {
```
jonhoo, line 71: We should document the safety requirements of `new_owned` here now that it's `unsafe`.

NobodyXu: Done!

src/native_mux_impl/stdio.rs
NobodyXu: > But isn't this the method that produces the fd that we send to the server?

It's actually `Stdio::{to_stdin, to_stdout, to_stderr}` that produces the fd (`PipeWrite` and `PipeRead`).
`Fd` holds onto them to keep them alive.

> If it isn't, then that implies that the fd is still going to be used going forward, in which case setting !non-blocking here may have unintended consequences.

I just realized that `Command::spawn` takes a mutable reference of `self`, so the `OwnedFd` of `Stdio` can indeed be reused.
Though they can only be used in `Command::spawn`, so setting !non-blocking should be fine.

src/child.rs
NobodyXu: I've updated the comment.

Fuuzetsu:
> @Fuuzetsu Can you test this PR to see if it fixed the bug for you please?

I checked with 0bb7ad321c09fd62ebaa41da94bf94e7505f0d86 and I'm now getting the stdout in full with `native-mux`. I only tried a couple of times.

NobodyXu:
> I checked with https://github.com/openssh-rust/openssh/commit/0bb7ad321c09fd62ebaa41da94bf94e7505f0d86 and I'm now getting the stdout in full with native-mux. I only tried a couple of times.

Thanks!
If you encountered the problem again when using this PR, feel free to post it here.

src/child.rs
```diff
@@ -62,6 +62,9 @@ macro_rules! delegate {
 /// available,`Stdio::piped()` should be passed to the corresponding method on
 /// [`Command`](crate::Command).
 ///
+/// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be sent to the
+/// remote process and `stdout` and `stderr` would eof immediately.
```
jonhoo, line 66:
```suggestion
/// remote process and `stdout` and `stderr` will yield EOF immediately.
```

src/native_mux_impl/stdio.rs
jonhoo: That feels like a foot-gun we're setting ourselves up for forgetting later and accidentally trying to use the FD in a tokio context at some later point in time, and it causing all sorts of hard-to-track-down problems. I'd definitely prefer for this to take `self`.

NobodyXu: > I'd definitely prefer for this to take self.

I can't take `self` here, since that would drop the `Fd::OwnedFd`.

jonhoo: You should be able to use [its implementation of `IntoRawFd`](https://doc.rust-lang.org/std/os/unix/io/struct.OwnedFd.html#method.into_raw_fd) to avoid that

NobodyXu: > You should be able to use [its implementation of IntoRawFd](https://doc.rust-lang.org/std/os/unix/io/struct.OwnedFd.html#method.into_raw_fd) to avoid that

But we still need to close that fd after sending them to ssh multiplex server.

Sending them over unix stream in linux is essentially `dup`ing them to another process.
If we don't close them, then they will be leaked.

jonhoo: I'm not sure I follow — once we send it to the multiplex server, why do we need to _also_ hold on to it and close it later? Won't the multiplex server close it at the appropriate time?

NobodyXu: > I'm not sure I follow — once we send it to the multiplex server, why do we need to also hold on to it and close it later? Won't the multiplex server close it at the appropriate time?

We send the fd to multiplex server over unix socket using `SCM_RIGHTS` ancillary data and it essentially acts as `dup` but across process boundary.

Sending a fd over would not automatically close it, that fd is still valid and can be used by the process.

[This blog](https://copyconstruct.medium.com/file-descriptor-transfer-over-unix-domain-sockets-dcbbf5b3b6ec) has more in-depth explanation of it.

NobodyXu: Hmmm, I believe we were talking past each other.

> why do we need to also hold on to it and close it later? Won't the multiplex server close it at the appropriate time?

After [the session is establied](https://github.com/openssh-rust/openssh/blob/5282f39b71d9750cc294d145c3d82ad57634ae0f/src/native_mux_impl/command.rs#L81), the [fds](https://github.com/openssh-rust/openssh/blob/5282f39b71d9750cc294d145c3d82ad57634ae0f/src/native_mux_impl/command.rs#L68) are still valid and we need to close them if `Stdio::pipe` is used as stdio.

jonhoo: Right, so we have something like this:

stdin_write_end, stdin_read_end, stdout_write_end, stdout_read_end (and ditto for stderr), each of which is a different fd. We send the stdin read end and the stdout write end to the multiplex server. We keep the stdin write end (so we can send stuff to stdin) and the stdout read end (so we can read what the multiplex server writes to stdout). But why do we need to keep the stdin read end and the stdout write end? Those _should_ be closed by the multiplex server when the remote process is done with them. We should never need to close those. We _will_ need to close the stdin write end (to signal no more input) and we _may_ close the stdout read end (to signal we don't care about any more of the remote process' output), but those are different FDs. Unless I'm missing something?

NobodyXu: > But why do we need to keep the stdin read end and the stdout write end?

We don't need it, that's why we need to close it to avoid leaking.
The whole purpose of enum `Fd` is to close these fds to avoid leaking.

> Those should be closed by the multiplex server when the remote process is done with them.

Yes.

> We should never need to close those.

Unfortunately we do, because sending them over unix socket using `SCM_RIGHTS` is essentailly `dup`ing them, so we still need to close the original one.

> We will need to close the stdin write end (to signal no more input) and we may close the stdout read end (to signal we don't care about any more of the remote process' output), but those are different FDs. Unless I'm missing something?

It's just how `SCM_RIGHTS` over unix socket work:

When you send a fd over unix socket using `SCM_RIGHTS` to another process, they **`dup`** the fds in another process's file descriptor table.

It does not close the fd in the process that sent it, you still need to close it manually.

To explain it in rust's term, sending fd over unix socket using `SCM_RIGHTS` takes a `BorrowedFd` not `OwnedFd`.
It takes the fd by reference not ownership so it's still valid after sending it over to the multiplex server.

jonhoo: Ah, okay, I see, so because it's dup'd by sending, we need to close it on our end. That seems reasonable, but that also implies to me that we should close it immediately after sending it. And this doesn't enforce that. One way to enforce that invariant and not have the possibility of accidentally doing the wrong thing is:
```rust
fn as_raw_fd_or_null_fd(self) -> Result<(RawFd, Option<OwnedFd>), Error> {
```
that way, the place we call this _has_ to ensure that the Fd will no longer be used anywhere, and is also given the `OwnedFd` so that it'll be dropped once the `RawFd` has been sent. Another alternative would be:
```rust
fn give_fd_to(self, f: impl FnOnce(RawFd) -> Result<(), Error>) -> Result<(), Error> {
```
though that may be a little more annoying if `f` needs to be `async`.

NobodyXu: > That seems reasonable, but that also implies to me that we should close it immediately after sending it.

That's what [`Command::spawn`](https://github.com/openssh-rust/openssh/blob/5282f39b71d9750cc294d145c3d82ad57634ae0f/src/native_mux_impl/command.rs#L64) is doing now.

It's closed immediately after being sent.

> `fn as_raw_fd_or_null_fd(self) -> Result<(RawFd, Option<OwnedFd>), Error> {`

This api is actually quite similar to `Stdio::{to_stdin, to_stdout, to_stderr}`.

The `Stdio::to_std*` apis returns `Result<(Fd, Option<Child*>), Error>`, then `Command::spawn` calls `as_raw_fd_or_null_fd` to establish an `[RawFd; 3]` array which is then sent to multiplex server using mux-client.

Once done, the `Fd`s would all get dropped.

jonhoo: Sure, but _this_ function is prone to misuse, because it changes the filedescriptor to non-blocking without also ensuring it'll never be used in tokio context again. I'd much prefer that we only change it to non-blocking just before we give it away, at a point where we can guarantee that it'll no longer be accessible to our tokio side of things.

NobodyXu: That makes sense, I would change that right away.

NobodyXu: I've changed the code so that now `set_blocking` is invoked in `Command::spawn`.
This makes it clear that the fd should not be reused and that it is indeed set to blocking mode before sending to the multiplex server.

src/native_mux_impl/command.rs
```diff
@@ -71,6 +71,10 @@ impl Command {
             stderr.as_raw_fd_or_null_fd()?,
         ];
 
+        for stdio in stdios {
+            set_blocking(stdio).map_err(Error::ChildIo)?;
+        }
```
jonhoo, line 76: This is much better, thank you! One last request: can we add a comment here along the lines of:

> Note that once we do this, these file descriptors (and the `Fd`s we got them from above) should no longer be used in any async context, as they'd start blocking. We give away the descriptors in `stdios` when we pass them to `open_new_session` (which doesn't use them in an async context), and the `Fd`s are dropped when this function returns, meaning no owned file descriptors we set to be blocking here can be used in an async context in the future.

jonhoo: Actually, writing the above made me realize that this isn't true for borrowed FDs, which include `StdioImpl::Fd` and, more importantly, `StdioImpl::Inherit`. This code, as written, will change the program's standard in and out to be blocking. Now, I _think_ that's okay for tokio given the not about a separate blocking thread [in their docs](https://docs.rs/tokio/latest/tokio/io/struct.Stdin.html), but it's probably unexpected for callers. And doubly so for borrowed FDs. I realize we may not have a choice, but I think this needs to be documented more explicitly!

jonhoo (changes requested):

src/native_mux_impl/command.rs
NobodyXu: It's ok for `StdioImpl::Fd` since they owned the fd and `Stdio` is only used in `Command::spawn`.

For `StdioImpl::Inherit`, that is a problem I did not consider, maybe I can skip remolving non-blocking flags specifically for inherit mode.

NobodyXu: Done!

I've added the comment and also skip `set_blocking` for `StdioImpl::Inherit`.

jonhoo: Hmm, is that true for `Fd::Borrowed` as well? It doesn't look like it is.

Also, why is it okay to skip this setting on the stdio fds? Won't the mux server still complain if those are non blocking too?

This also makes me skeptical of [this `impl`](https://github.com/openssh-rust/openssh/blob/master/src/stdio.rs#L49), which I think means any raw fd (including one that isn't owned) can be turned into one that *looks* owned (which is a problem).

NobodyXu: > Hmm, is that true for Fd::Borrowed as well? It doesn't look like it is.

I only skipped `set_blocking` for inherit.

> Also, why is it okay to skip this setting on the stdio fds? Won't the mux server still complain if those are non blocking too?

Well, that's indeed a problem, but I think it's bad to change fd 0, 1, 2 since they are considered as global state?

> This also makes me skeptical of [this impl](https://github.com/openssh-rust/openssh/blob/master/src/stdio.rs#L49), which I think means any raw fd (including one that isn't owned) can be turned into one that looks owned (which is a problem).

Hmmm, yeah.
Now looking at that it's also a problem.

Maybe we should only `set_blocking` for the `PipeRead`/`PipeWrite` we explicitly created?

NobodyXu: @jonhoo I've taken the route where we only call `set_blocking` fd we are 100% is owned, which are:
 - fds constructed from `OwnedFd`
 - fds constructed from other owned type, such as `PipeWrite`/`PipeRead`, `File`, `ChildStd*`.
 - `StdioImpl::Pipe`

For `StdioImpl::Inherit` and fds constructed from `FromRawFd::from_raw_fd`, `set_blocking` is not called and instead I added documentation for them.

src/stdio.rs
```diff
@@ -40,13 +40,23 @@ impl Stdio {
...
+/// FromRawFd takes ownership of the fd passed in
+/// and closes the fd on drop.
+///
+/// NOTE that the fd must be in blocking mode, otherwise
+/// ssh might not flush all output since it considers
+/// (`EAGAIN`/`EWOULDBLOCK`) as an error
 impl FromRawFd for Stdio {
     unsafe fn from_raw_fd(fd: RawFd) -> Self {
-        Self(StdioImpl::Fd(OwnedFd::from_raw_fd(fd)))
+        Self(StdioImpl::Fd(OwnedFd::from_raw_fd(fd), false))
```
jonhoo, line 59: Feels like this should really just be a `BorrowedFd` at this point, no?

jonhoo: And then we could also add a `From<OwnedFd>` impl for things that are actually owned.

src/native_mux_impl/stdio.rs
```diff
@@ -28,25 +29,74 @@ fn get_null_fd() -> Result<RawFd, Error> {
...
+    /// # Safety
+    ///
+    /// `T::into_raw_fd` must return a valid fd and transfers
+    /// the ownershipt of it.
+    unsafe fn new_owned<T: IntoRawFd>(fd: T) -> Result<Self, Error> {
+        let raw_fd = fd.into_raw_fd();
+        // Create owned_fd so that the fd will be closed on error
+        let owned_fd = OwnedFd::from_raw_fd(raw_fd);
+
+        set_blocking(raw_fd).map_err(Error::ChildIo)?;
```
jonhoo, line 72: I still think we should keep this call to just before we _need_ it to be blocking, rather than the moment we get one. For example, as written, I _think_ this would also set blocking on the ends of the pipe that we end up keeping? I think it'd be better for this call to go in `to_stdin`/`to_output`.

jonhoo (changes requested):

src/stdio.rs
NobodyXu: > Feels like this should really just be a BorrowedFd at this point, no?

`BorrowedFd` has a lifetime parameter `'fd`, using it would certainly break the apis.

> And then we could also add a From<OwnedFd> impl for things that are actually owned.

Problem is people already uses `FromRawFd` and they probably expects the fd to be closed.

src/native_mux_impl/stdio.rs
NobodyXu: I've moved all the calls to `to_stdin`/`to_output` and also added doc on why read/write is set to blocking.

src/native_mux_impl/stdio.rs
```diff
@@ -28,25 +29,73 @@ fn get_null_fd() -> Result<RawFd, Error> {
...
+    ///
+    /// `T::into_raw_fd` must return a valid fd and transfers
+    /// the ownershipt of it.
+    unsafe fn new_owned<T: IntoRawFd>(fd: T) -> Result<Self, Error> {
+        let raw_fd = fd.into_raw_fd();
+        Ok(Fd::Owned(OwnedFd::from_raw_fd(raw_fd)))
+    }
+}
+
+impl TryFrom<PipeRead> for Fd {
```
jonhoo, line 77: I think we should make these private functions instead of implementing the trait, because otherwise `PipeRead` and `PipeWrite` become part of our public API whereas I don't think they currently are.

src/stdio.rs
jonhoo: That's a fair point. It does mean that the API is kind of sharp in multiple ways.

Maybe mark this whole `impl` as `#[deprecated]` and add a new inherent method that explicitly documents that it will both close the FD _and_ mark it as blocking? That way, existing code will keep working, but new adopters will be able to get the correct/expected behavior. And then in the next major release, we can remove this impl `FromRawFd` in favor of `From<OwnedFd>`.

src/native_mux_impl/stdio.rs
NobodyXu: Emmm they are already part of the public api.

These TryFrom impl is present in 0.9.0, but we can remove that in the next major release.

jonhoo: I'm confused — they show as added in this PR? Where did they live before? If they truly exited before, we should at least mark them as deprecated.

src/stdio.rs
NobodyXu: I've marked the `FromRawFd` impl for `Stdio` as `#[deprecated]`, but it's currently "useless" due to https://github.com/rust-lang/rust/issues/39935

> add a new inherent method that explicitly documents that it will both close the FD and mark it as blocking?

And I've added new method `Stdio::from_raw_fd_owned` for that.

src/native_mux_impl/stdio.rs
NobodyXu: They indeed [exists](https://github.com/openssh-rust/openssh/blob/5282f39b71d9750cc294d145c3d82ad57634ae0f/src/stdio.rs#L86) before this PR.

I've added too many trait impl and back then I didn't think of the effect of it on the API.

I can certainly mark it as deprecated, though due to https://github.com/rust-lang/rust/issues/39935 it will be warned as "useless".

jonhoo: Ah, those impls are for `Stdio`, not for `Fd`. And, looking at it some more, since `Fd` isn't a public type, this particular impl isn't a problem. Those impls are, though not much to do about those now :sweat_smile:

jonhoo (approved):

jonhoo:
Oh, can you update the changelog as well?

src/native_mux_impl/stdio.rs
NobodyXu: Done!

I've also updated their doc to reflect that, since `#[deprecate]` does not change the doc at all.

NobodyXu:
> Oh, can you update the changelog as well?

I've updated the changelog.

jonhoo (approved):

NobodyXu:
P.S. codecov CI seems to fail due to the invalid token:

```
[2022-11-07T01:01:09.453Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v2/dist/codecov' failed with exit code 255
```

jonhoo:
Oops, I forgot to squash.. Oh well!

jonhoo:
Published in 0.9.8 :tada:

jonhoo:
I think that CI failure is probably temporary

NobodyXu:
> I think that CI failure is probably temporary

That indeeds seems so since it worked again on master branch.

PR #109, opened 2022-12-21
Feature: Support `ssh-agent`

### LINKED ISSUE #108: SSH_AUTH_SOCK environment variable should be passed to SSH
Right now it seems that `openssh::Session::connect_mux` does not pass `SSH_AUTH_SOCK` environment variable to the child SSH process, breaking all setups that depend on an SSH agent.

This is exhibited by getting a PermissionDenied error when trying to establish a connection ` Connect(Custom { kind: PermissionDenied, error: "pin required\r\nuser@example.com: Permission denied (publickey)." })'` while the same host does not prompt for anything when using `ssh` standalone.

NobodyXu:
Fixed #108, implemented new API `SessionBuilder::ssh_auth_sock

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -208,6 +208,9 @@ impl SessionBuilder {
     /// This is equivalent to:
     ///
     /// ```
+    /// use openssh::SessionBuilder;
+    /// use std::{env, path::PathBuf};
+    ///
     /// let mut builder = SessionBuilder::default();
     ///
     /// if let Some(ssh_auth_sock) = env::var_os("SSH_AUTH_SOCK") {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,6 +1,7 @@
 use super::{Error, Session};
 
 use std::borrow::Cow;
+use std::env;
 use std::ffi::OsString;
 use std::fs;
 use std::iter::IntoIterator;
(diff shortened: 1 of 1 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/109)

src/builder.rs
```diff
@@ -189,6 +192,38 @@ impl SessionBuilder {
...
+    /// use openssh::SessionBuilder;
+    /// use std::{env, path::PathBuf};
+    ///
+    /// let mut builder = SessionBuilder::default();
+    ///
+    /// if let Some(ssh_auth_sock) = env::var_os("SSH_AUTH_SOCK") {
+    ///     builder.ssh_auth_sock(PathBuf::from(ssh_auth_sock));
+    /// }
+    /// ```
+    pub fn inherit_ssh_auth_sock(&mut self) -> &mut Self {
```
jonhoo, line 220: I'm confused by this. We never call `env_clear` (or is it `clear_env`?) on the `CommandBuilder`, so environment variables should apply to the sub-process when we execute it regardless. Why have a dedicated method for this behavior when it's already what'll be happening?

NobodyXu: Oops I forgot this when I saw the comment https://github.com/openssh-rust/openssh/issues/108#issue-1504346738 

I suppose I should remove this method then, since @TaaviE apparently has encountered other issues.

NobodyXu: Done.

src/builder.rs
```diff
@@ -341,6 +355,10 @@ impl SessionBuilder {
             init.arg("-o").arg(format!("Compression={}", arg));
         }
 
+        if let Some(ssh_auth_sock) = self.ssh_auth_sock.as_deref() {
+            init.env("SSH_AUTH_SOCK", ssh_auth_sock);
```
jonhoo, line 359: Can this not be passed using an argument instead?

src/builder.rs
```diff
@@ -1,6 +1,7 @@
 use super::{Error, Session};
 
 use std::borrow::Cow;
+use std::env;
```
jonhoo, line 4: This is unused now

jonhoo (changes requested):

src/builder.rs
NobodyXu: Done!

src/builder.rs
NobodyXu: I checked the man page of `ssh` and googled it but didn't find anything

jonhoo (approved):
How weird that it _must_ be an environment variable. But fair enough!

jonhoo:
Ah, would you mind updating the changelog as well?

NobodyXu:
> Ah, would you mind updating the changelog as well?

Sure, I forgot about that

PR #112, opened 2023-01-31
Allow conversion of `std::process::Command` into a `openssh::Command` given a `openssh::Session`.

aalekhpatel07:
The `std::process::Command` and the `openssh::Command` APIs are fairly similar, so I wonder if it is reasonable to allow an arbitrary `std::process::Command` to be run over ssh.

#### Example

```rust
use std::process;
use openssh::{OverSsh, Session, KnownHosts};

#[tokio::main]
pub async fn main() {
    let session = Session::connect_mux("me@ssh.example.com", KnownHosts::Strict).await.unwrap();

    let ls = process::Command::new("ls")
        .arg("-l")
        .arg("-h")
        .over_ssh(&session)
        .expect("no env vars or current_dir are specified")
        .output()
        .await
        .unwrap();

    println!("{}", String::from_utf8_lossy(&ls.stdout));
}


```

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -94,7 +94,7 @@ impl OverSsh for std::process::Command {
         
         let args = self
             .get_args()
-            .map(|arg| escape(arg));
+            .map(escape);
         command.raw_args(args);
         command
     }
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,3 +1,5 @@
+use crate::escape;
+
 use super::stdio::TryFromChildIo;
 use super::RemoteChild;
 use super::Stdio;
@@ -5,6 +7,7 @@ use super::{Error, Session};
 
 use std::borrow::Cow;
 use std::ffi::OsStr;
+use std::os::unix::prelude::OsStrExt;
 use std::process;
 
 #[derive(Debug)]
@@ -57,15 +60,10 @@ pub trait OverSsh {
     /// **Note**: The command to be executed on the remote machine does not include 
     /// any environment variables or the current directory. They must be
     /// set explicitly, if desired.
-    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
-}
-
-impl OverSsh for std::process::Command {
-    /// Given a session, convert a `std::process::Command` into an `openssh::Command` 
-    /// that can be executed over that session.
-    /// **Note**: The command to be executed on the remote machine does not include 
-    /// any environment variables or the current directory. They must be
-    /// set explicitly, if desired.
+    /// 
+    /// # Example
+    /// Consider the implementation of `OverSsh` for `std::process::Command`:
+    /// 
     /// ```no_run
     /// # #[tokio::main(flavor = "current_thread")]
     /// async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -87,42 +85,26 @@ impl OverSsh for std::process::Command {
     /// }
     /// 
     /// ```
+    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
+}
+
+impl OverSsh for std::process::Command {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        let mut command = session.command(self.get_program().to_string_lossy());
-        command.raw_args(self.get_args());
+        let program_escaped = escape(self.get_program().as_bytes());
+        let mut command = session.command(Cow::Borrowed(program_escaped.as_str()));
+        
+        self
+        .get_args()
+        .for_each(|arg| {
+            let arg_escaped = escape(arg.as_bytes());
+            command.arg(Cow::Borrowed(arg_escaped.as_str()));
+        });
         command
     }
 }
 
 
 impl OverSsh for tokio::process::Command {
-    /// Given a session, convert a `tokio::process::Command` into an `openssh::Command` 
-    /// that can be executed over that session.
-    /// 
-    /// **Note**: The command to be executed on the remote machine does not include 
-    /// any environment variables or the current directory. They must be
-    /// set explicitly, if desired.
-    /// ```no_run
-    /// # #[tokio::main(flavor = "current_thread")]
-    /// # async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    ///     use tokio::process::Command;
-    ///     use openssh::{Session, KnownHosts, OverSsh};
-
-    ///     let session = Session::connect_mux("me@ssh.example.com", KnownHosts::Strict).await?;
-    ///     let ls = 
-    ///         Command::new("ls")
-    ///         .arg("-l")
-    ///         .arg("-a")
-    ///         .arg("-h")
-    ///         .over_session(&session)
-    ///         .output()
-    ///         .await?;
-    /// 
-    ///     assert!(String::from_utf8(ls.stdout).unwrap().contains("total"));
-    /// #   Ok(())
-    /// # }
-    /// 
-    /// ```
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
         self.as_std().over_session(session)
     }
@@ -130,15 +112,6 @@ impl OverSsh for tokio::process::Command {
 
 
 impl<S> OverSsh for &S
-where
-    S: OverSsh
-{
-    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        <S as OverSsh>::over_session(self, session)
-    }
-}
-
-impl<S> OverSsh for &mut S
 where
     S: OverSsh
 {
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,4 +1,4 @@
-use crate::escape;
+use crate::escape::escape;
 
 use super::stdio::TryFromChildIo;
 use super::RemoteChild;
@@ -7,7 +7,6 @@ use super::{Error, Session};
 
 use std::borrow::Cow;
 use std::ffi::OsStr;
-use std::os::unix::prelude::OsStrExt;
 use std::process;
 
 #[derive(Debug)]
@@ -90,15 +89,13 @@ pub trait OverSsh {
 
 impl OverSsh for std::process::Command {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        let program_escaped = escape(self.get_program().as_bytes());
-        let mut command = session.command(Cow::Borrowed(program_escaped.as_str()));
+        let program_escaped: Cow<'_, OsStr> = escape(self.get_program());
+        let mut command = session.raw_command(program_escaped);
         
-        self
-        .get_args()
-        .for_each(|arg| {
-            let arg_escaped = escape(arg.as_bytes());
-            command.arg(Cow::Borrowed(arg_escaped.as_str()));
-        });
+        let args = self
+            .get_args()
+            .map(|arg| escape(arg));
+        command.raw_args(args);
         command
     }
 }
--- a/src/command.rs
+++ b/src/command.rs
@@ -52,16 +52,23 @@ macro_rules! delegate {
 }
 
 /// If a command is `OverSsh` then it can be executed over an SSH session.
+/// 
 /// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
 pub trait OverSsh {
-    /// Given a session, return a command that can be executed over that session.
+    /// Given an ssh session, return a command that can be executed over that ssh session.
     ///
-    /// **Note**: The command to be executed on the remote machine does not include
-    /// any environment variables or the current directory. They must be
-    /// set explicitly, if desired.
-    ///
-    /// # Example
-    /// Consider the implementation of `OverSsh` for `std::process::Command`:
+    /// ### Notes
+    /// 
+    /// The command to be executed on the remote machine should not explicitly
+    /// set environment variables or the current working directory. It errors if the source command 
+    /// has environment variables or a current working directory set, since `openssh` doesn't (yet) have
+    /// a method to set environment variables and `ssh` doesn't support setting a current working directory
+    /// outside of `bash/dash/zsh` (which is not always available).
+    /// 
+    /// ###  Examples
+    /// 
+    /// 1. Consider the implementation of `OverSsh` for `std::process::Command`. Let's build a
+    /// `ls -l -a -h` command and execute it over an SSH session.
     ///
     /// ```no_run
     /// # #[tokio::main(flavor = "current_thread")]
@@ -75,7 +82,7 @@ pub trait OverSsh {
     ///         .arg("-l")
     ///         .arg("-a")
     ///         .arg("-h")
-    ///         .over_session(&session)
+    ///         .over_ssh(&session)?
     ///         .output()
     ///         .await?;
     ///
@@ -84,23 +91,54 @@ pub trait OverSsh {
     /// }
     ///
     /// ```
-    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
+    /// 2. Building a command with environment variables or a current working directory set will
+    /// results in an error.
+    /// 
+    /// ```no_run
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    ///     use std::process::Command;
+    ///     use openssh::{Session, KnownHosts, OverSsh};
+    ///
+    ///     let session = Session::connect_mux("me@ssh.example.com", KnownHosts::Strict).await?;
+    ///     let echo =
+    ///         Command::new("echo")
+    ///         .arg("$MY_ENV_VAR")
+    ///         .over_ssh(&session);
+    ///     assert_matches!(echo, Err(openssh::Error::CommandHasEnv));
+    ///
+    /// #   Ok(())
+    /// }
+    ///
+    /// ```
+    fn over_ssh<'session>(&self, session: &'session Session) -> Result<crate::Command<'session>, crate::Error>;
 }
 
 impl OverSsh for std::process::Command {
-    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
+    fn over_ssh<'session>(&self, session: &'session Session) -> Result<Command<'session>, crate::Error> {
+
+        // I'd really like `!self.get_envs().is_empty()` here, but that's
+        // behind a `exact_size_is_empty` feature flag.
+        if self.get_envs().len() > 0 {
+            return Err(crate::Error::CommandHasEnv);
+        }
+
+        if self.get_current_dir().is_some() {
+            return Err(crate::Error::CommandHasCwd);
+        }
+
         let program_escaped: Cow<'_, OsStr> = escape(self.get_program());
         let mut command = session.raw_command(program_escaped);
 
         let args = self.get_args().map(escape);
         command.raw_args(args);
-        command
+        Ok(command)
     }
 }
 
 impl OverSsh for tokio::process::Command {
-    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        self.as_std().over_session(session)
+    fn over_ssh<'session>(&self, session: &'session Session) -> Result<Command<'session>, crate::Error> {
+        self.as_std().over_ssh(session)
     }
 }
 
@@ -108,37 +146,20 @@ impl<S> OverSsh for &S
 where
     S: OverSsh,
 {
-    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        <S as OverSsh>::over_session(self, session)
+    fn over_ssh<'session>(&self, session: &'session Session) -> Result<Command<'session>, crate::Error> {
+        <S as OverSsh>::over_ssh(self, session)
     }
 }
 
-impl<S> OverSsh for Box<S>
+impl<S> OverSsh for &mut S
 where
     S: OverSsh,
 {
-    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        <S as OverSsh>::over_session(self, session)
+    fn over_ssh<'session>(&self, session: &'session Session) -> Result<Command<'session>, crate::Error> {
+        <S as OverSsh>::over_ssh(self, session)
     }
 }
 
-impl<S> OverSsh for std::rc::Rc<S>
-where
-    S: OverSsh,
-{
-    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        <S as OverSsh>::over_session(self, session)
-    }
-}
-
-impl<S> OverSsh for std::sync::Arc<S>
-where
-    S: OverSsh,
-{
-    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
-        <S as OverSsh>::over_session(self, session)
-    }
-}
 
 /// A remote process builder, providing fine-grained control over how a new remote process should
 /// be spawned.
--- a/src/command.rs
+++ b/src/command.rs
@@ -55,14 +55,14 @@ macro_rules! delegate {
 /// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
 pub trait OverSsh {
     /// Given a session, return a command that can be executed over that session.
-    /// 
-    /// **Note**: The command to be executed on the remote machine does not include 
+    ///
+    /// **Note**: The command to be executed on the remote machine does not include
     /// any environment variables or the current directory. They must be
     /// set explicitly, if desired.
-    /// 
+    ///
     /// # Example
     /// Consider the implementation of `OverSsh` for `std::process::Command`:
-    /// 
+    ///
     /// ```no_run
     /// # #[tokio::main(flavor = "current_thread")]
     /// async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -70,7 +70,7 @@ pub trait OverSsh {
     ///     use openssh::{Session, KnownHosts, OverSsh};
     ///
     ///     let session = Session::connect_mux("me@ssh.example.com", KnownHosts::Strict).await?;
-    ///     let ls = 
+    ///     let ls =
     ///         Command::new("ls")
     ///         .arg("-l")
     ///         .arg("-a")
@@ -78,11 +78,11 @@ pub trait OverSsh {
     ///         .over_session(&session)
     ///         .output()
     ///         .await?;
-    /// 
+    ///
     ///     assert!(String::from_utf8(ls.stdout).unwrap().contains("total"));
     /// #   Ok(())
     /// }
-    /// 
+    ///
     /// ```
     fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
 }
@@ -91,15 +91,12 @@ impl OverSsh for std::process::Command {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
         let program_escaped: Cow<'_, OsStr> = escape(self.get_program());
         let mut command = session.raw_command(program_escaped);
-        
-        let args = self
-            .get_args()
-            .map(escape);
+
+        let args = self.get_args().map(escape);
         command.raw_args(args);
         command
     }
 }
-
 
 impl OverSsh for tokio::process::Command {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
@@ -107,10 +104,9 @@ impl OverSsh for tokio::process::Command {
     }
 }
 
-
 impl<S> OverSsh for &S
 where
-    S: OverSsh
+    S: OverSsh,
 {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
         <S as OverSsh>::over_session(self, session)
@@ -119,7 +115,7 @@ where
 
 impl<S> OverSsh for Box<S>
 where
-    S: OverSsh
+    S: OverSsh,
 {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
         <S as OverSsh>::over_session(self, session)
@@ -128,7 +124,7 @@ where
 
 impl<S> OverSsh for std::rc::Rc<S>
 where
-    S: OverSsh
+    S: OverSsh,
 {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
         <S as OverSsh>::over_session(self, session)
@@ -137,7 +133,7 @@ where
 
 impl<S> OverSsh for std::sync::Arc<S>
 where
-    S: OverSsh
+    S: OverSsh,
 {
     fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
         <S as OverSsh>::over_session(self, session)
--- a/src/command.rs
+++ b/src/command.rs
@@ -103,9 +103,10 @@ pub trait OverSsh {
     ///     let session = Session::connect_mux("me@ssh.example.com", KnownHosts::Strict).await?;
     ///     let echo =
     ///         Command::new("echo")
+    ///         .env("MY_ENV_VAR", "foo")
     ///         .arg("$MY_ENV_VAR")
     ///         .over_ssh(&session);
-    ///     assert_matches!(echo, Err(openssh::Error::CommandHasEnv));
+    ///     assert!(matches!(echo, Err(openssh::Error::CommandHasEnv)));
     ///
     /// #   Ok(())
     /// }
--- a/src/command.rs
+++ b/src/command.rs
@@ -49,6 +49,41 @@ macro_rules! delegate {
     }};
 }
 
+/// If a command is `OverSsh` then it can be executed over an SSH session.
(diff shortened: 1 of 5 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/112)

src/command.rs
```diff
@@ -49,6 +49,42 @@ macro_rules! delegate {
...
 }
 
+/// If a command is `OverSsh` then it can be executed over an SSH session.
+/// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
+pub trait OverSsh {
+    /// Given a session, return a command that can be executed over that session.
+    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
+}
+
+impl OverSsh for std::process::Command {
```
NobodyXu, line 59: We could also implement this for `tokio::process::Command`.

src/command.rs
```diff
@@ -49,6 +49,42 @@ macro_rules! delegate {
     }};
 }
 
+/// If a command is `OverSsh` then it can be executed over an SSH session.
+/// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
+pub trait OverSsh {
+    /// Given a session, return a command that can be executed over that session.
+    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
+}
+
```
NobodyXu, line 58: I think it will be also great to have some stub implementation for `&T`, `Box<T>`, `Arc<T>` and `Rc<T>` like this:

```
impl<T: OverSsh> OverSsh for &T {
    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session> {
        <T as OverSsh>::over_session(session)
    }
}
```

src/command.rs
```diff
@@ -49,6 +49,42 @@ macro_rules! delegate {
...
+    ///     .output()
+    ///     .await?;
+    /// 
+    /// assert!(String::from_utf8(ls.stdout).unwrap().contains("total"));
+    /// 
+    /// 
+    /// ```
+    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
+
+        let mut command = session.raw_command(self.get_program());
```
NobodyXu, line 82: I think we should use `Session::command` here to escape the program, since the ssh protocol essentially concats program and args into one string `$program $arg0 $arg1 ...` and pass it to shell.

src/command.rs
```diff
@@ -49,6 +49,42 @@ macro_rules! delegate {
...
+        command.raw_args(self.get_args());
```
NobodyXu, line 83: Same here, I think `Command::args` is more appropriate here.

src/command.rs
```diff
@@ -49,6 +49,42 @@ macro_rules! delegate {
...
+/// If a command is `OverSsh` then it can be executed over an SSH session.
+/// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
+pub trait OverSsh {
+    /// Given a session, return a command that can be executed over that session.
+    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
+}
+
+impl OverSsh for std::process::Command {
+    /// Given a session, convert a std::process::Command into an `openssh::Command` 
+    /// that can be executed over that session.
```
NobodyXu, line 61: We should also document that this does not include any environment variable and the current dir.

NobodyXu (changes requested):
Thanks!

I really like the design and the implementation is simple, but there are a few places where it can be improved.

aalekhpatel07:
wow thank you for your kind words and a quick review. I'll clean it up right away! Thanks for your time.

src/command.rs
NobodyXu: For `Box<T>`, `Arc<T>` and `Rc<T>`, we could perhaps use [`AsRef`](https://doc.rust-lang.org/std/convert/trait.AsRef.html) trait, which is more extensible and will work for types defined in other crates, such as `triomphe`'s `Arc`, though I'm really not sure whether this will work, so it would need some testing:

```
impl<T: AsRef<U>, U: OverSsh> OverSsh for T {
    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session> {
        self.as_ref().over_session(session)
    }
}
```

For `&T`, we might still need to manually implement for them.

src/command.rs
```diff
@@ -49,6 +49,42 @@ macro_rules! delegate {
...
+    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
+}
+
+impl OverSsh for std::process::Command {
+    /// Given a session, convert a `std::process::Command` into an `openssh::Command` 
+    /// that can be executed over that session.
+    /// ```no_run
+    /// use std::process::Command;
+    /// use openssh::{Session, KnownHosts, OverSsh};
+
```
NobodyXu, line 65:
```suggestion
    /// # #[tokio::main(flavor = "current_thread")]
    /// # async fn main() -> Result<(), Box<dyn Error>> {
```

NobodyXu: Inside the code block, anything starts with `#` will be hidden, so you can use `#[tokio::main]` to fix the error here.

src/command.rs
```diff
@@ -49,6 +49,42 @@ macro_rules! delegate {
...
+    ///     .arg("-l")
+    ///     .arg("-a")
+    ///     .arg("-h")
+    ///     .over_session(&session)
+    ///     .output()
+    ///     .await?;
+    /// 
+    /// assert!(String::from_utf8(ls.stdout).unwrap().contains("total"));
+    /// 
+    ///
```
NobodyXu, line 78:
```suggestion
    /// #    Ok(())
    /// # }
```

NobodyXu (changes requested):

src/command.rs
aalekhpatel07: hmm I'm not sure I fully understand this. As it stands, `self.get_program()` returns an `&OsStr` but `Session::command` expects `Into<Cow<'a, str>>`. A meaningful way to get that is to `self.get_program().to_string_lossy()` but I'm a little worried that changes the semantics here. please lmk if I misunderstood it.

NobodyXu: Oh I didn't realize the difference between typing.

Yeah that's certainly an issue, but I still think it's necessary to escape them, e.g. if the program contains `'` or `!`, then it will be interpreted by the shell instead of as part of the program name.

We could use [`OsStrExt::as_bytes`](https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStrExt.html#tymethod.as_bytes) then escapes them in the same way as [`shell_escape::unix::escape`](https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#101-120).

src/command.rs
aalekhpatel07: No dice with the type checker. Seems like `U` isn't sufficiently constrained. Even this doesn't work:

```rust
impl<T: Into<U>, U: Foo> Foo for T {
    ...
}
```

I feel like I've tried to write something similar before and come across it. I'm not sure I know what its called though.

src/command.rs
aalekhpatel07: There are still some typing incompatibilities as `shell_escape::unix::escape` expects a `Cow<str>`. I tried to implement a variant for `Cow<OsString> -> Cow<OsString>` but I couldn't find a way to interleave special (i.e. `[^\!']`) characters with the non-special ones. 

Hoping you could suggest a fix for what you have in mind.

src/command.rs
aalekhpatel07: same story as with `Session::command` (i.e. expected `Cow<str>` but provided `Cow<OsStr>`).

src/command.rs
NobodyXu: Well, that's not unexpected...

For now let's go with manually impl for `&T`, `Box<T>`, `Arc<T>` and `Rc<T>`

src/command.rs
```diff
@@ -49,6 +49,131 @@ macro_rules! delegate {
...
+    /// that can be executed over that session.
+    /// **Note**: The command to be executed on the remote machine does not include 
+    /// any environment variables or the current directory. They must be
+    /// set explicitly, if desired.
+    /// ```no_run
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    ///     use std::process::Command;
+    ///     use openssh::{Session, KnownHosts, OverSsh};
+
```
NobodyXu, line 74:
```suggestion
    ///
```

Having a blank line like this would not show up in the example on docs.rs

It has to be prefixed with `///`.

src/command.rs
```diff
@@ -49,6 +49,131 @@ macro_rules! delegate {
...
+}
+
+impl<S> OverSsh for &mut S
+where
+    S: OverSsh
+{
+    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
+        <S as OverSsh>::over_session(self, session)
+    }
+}
```
NobodyXu, line 148: I don't think we need this as `impl<S> OverSsh for &S` should be enough and `rustc` has auto deref rules that should take care for this.

src/command.rs
```diff
@@ -49,6 +49,131 @@ macro_rules! delegate {
...
+    ///         .arg("-a")
+    ///         .arg("-h")
+    ///         .over_session(&session)
+    ///         .output()
+    ///         .await?;
+    /// 
+    ///     assert!(String::from_utf8(ls.stdout).unwrap().contains("total"));
+    /// #   Ok(())
+    /// # }
+    ///
```
NobodyXu, line 124: Let's move the doc to the definition of `OverSsh` trait, since users will check it first and it will avoid the dup doc for both impl.

NobodyXu (changes requested):

src/command.rs
NobodyXu: You can use [`OsStrExt::as_bytes`](https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStrExt.html#tymethod.as_bytes) to convert `OsStr` to `&[u8]` first, then you can implement [`shell_escape::unix::escape`](https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#101-120) but for bytes `&[u8]` instead of `&str`.

NobodyXu: I just recall there is plan to support windows, but it will take probably at least another 3-6 months before it is even ready for review (I am working some other projects right now).

So this is probably fine.
For windows, we would have to use some different implementations.

NobodyXu:
Ehhh we now have a rustc ICE.

@aalekhpatel07 Would you like to open an issue in rust-lang/rust?

aalekhpatel07:
> Ehhh we now have a rustc ICE.
> 
> @aalekhpatel07 Would you like to open an issue in rust-lang/rust?

Haha for sure. Any preference for a name?

NobodyXu:
@aalekhpatel07 Looks like some of the ICEs happens in mir borrow checking, so perhaps "mir borrow checking ICE in openssh-rust" will be a good name.

Also I recommend you to create a separate branch, containing the commit causing the ICEs just so it's easier for others to reproduce the ICEs.

src/command.rs
aalekhpatel07: that makes sense. lets go with this for now.

aalekhpatel07:
I believe the new `escape(s: &[u8]) -> String` should do the job for us. lmk

src/command.rs
```diff
@@ -49,6 +52,101 @@ macro_rules! delegate {
...
+    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
+        let program_escaped = escape(self.get_program().as_bytes());
+        let mut command = session.command(Cow::Borrowed(program_escaped.as_str()));
+        
+        self
+        .get_args()
+        .for_each(|arg| {
+            let arg_escaped = escape(arg.as_bytes());
+            command.arg(Cow::Borrowed(arg_escaped.as_str()));
+        });
```
NobodyXu, line 101:
```suggestion
        let args = self
            .get_args()
            .map(|arg| {
                let arg_escaped = escape(arg.as_bytes());
                Cow::Borrowed(arg_escaped.as_str())
            });
        command.args(args);
```

src/escape.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Escape characters that may have special meaning in a shell, including spaces.
+/// 
+/// **Note**: This function is an adaptation of [`shell-escape::unix::escape`].
+/// This function exists only for type compatibility and the implementation is
+/// almost exactly the same as [`shell-escape::unix::escape`].
+/// 
+/// [`shell-escape::unix::escape`]: https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#101
+/// 
+pub fn escape(s: &[u8]) -> String {
```
NobodyXu, line 23: Let's make this easier to use and easier to port to windows by accept `&OsStr`.

Also, I'd like the return type to be `Cow<'_, OsStr>`, not `String`.

Otherwise what's the point of using `OsStr`?
We could just use `OsStr::to_string_lossy`.

```suggestion
pub fn escape(s: &OsStr) -> Cow<'_, OsStr> {
```

src/escape.rs
```diff
@@ -0,0 +1,81 @@
...
+    let all_whitelisted = s.iter().all(|x| whitelisted(*x as char));
+
+    if !s.is_empty() && all_whitelisted {
+        // All bytes are whitelisted and valid single-byte UTF-8, 
+        // so we can build the original string and return as is.
+        return String::from_utf8(s.to_vec()).unwrap();
```
NobodyXu, line 29: This is just wrong, there's no guarantee that `OsStr` contains utf-8 characters.

src/lib.rs
```diff
@@ -167,7 +167,10 @@ mod builder;
 pub use builder::{KnownHosts, SessionBuilder};
 
 mod command;
-pub use command::Command;
+pub use command::{Command, OverSsh};
+
+mod escape;
+pub use escape::escape;
```
NobodyXu, line 173: No, let's not export this.
It should not be part of the public API.

NobodyXu: Actually, I'd like this to be a separate crate, but that is out of the scope of this PR.

Let's just make this crate-internal API.

src/escape.rs
```diff
@@ -0,0 +1,81 @@
+//! Escape characters that may have special meaning in a shell, including spaces.
+//! This is a modified version of the [`shell-escape::unix`] module of [`shell-escape`] crate.
+//! 
+//! [`shell-escape`]: https://crates.io/crates/shell-escape
+//! [`shell-escape::unix`]: https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#101
+
+
+fn whitelisted(ch: char) -> bool {
```
NobodyXu, line 8: Let's take byte here:

```suggestion
fn whitelisted(byte: u8) -> bool {
```

src/escape.rs
```diff
@@ -0,0 +1,81 @@
...
+pub fn escape(s: &[u8]) -> String {
+    let all_whitelisted = s.iter().all(|x| whitelisted(*x as char));
+
+    if !s.is_empty() && all_whitelisted {
+        // All bytes are whitelisted and valid single-byte UTF-8, 
+        // so we can build the original string and return as is.
+        return String::from_utf8(s.to_vec()).unwrap();
+    }
+
+    let mut escaped = String::with_capacity(s.len() + 2);
```
NobodyXu, line 32: Same here, should use `Vec` and then use [`std::os::unix::ffi::OsStringExt::from_vec`](https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStringExt.html#tymethod.from_vec) to convert it to `OsStr`.

NobodyXu (changes requested):

NobodyXu:
> (I'm sorry I think I accidentally removed the review request for jonhoo).

No worries, I've re-requested it.

NobodyXu:
Aha, the CI reliably failed with ICEs from `rustc`.

src/escape.rs
aalekhpatel07: that's my bad. I was mistaken in assuming that each of those bytes individually representing whitelisted chars implies they are to be interpreted as single byte utf-8 chars.

aalekhpatel07:
okay how about these changes? (thank you for being patient.)

src/command.rs
```diff
@@ -49,6 +51,99 @@ macro_rules! delegate {
...
+}
+
+impl OverSsh for std::process::Command {
+    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
+        let program_escaped: Cow<'_, OsStr> = escape(self.get_program());
+        let mut command = session.raw_command(program_escaped);
+        
+        let args = self
+            .get_args()
+            .map(|arg| escape(arg));
```
NobodyXu, line 97:
```suggestion
            .map(escape);
```

src/lib.rs
```diff
@@ -167,7 +167,9 @@ mod builder;
 pub use builder::{KnownHosts, SessionBuilder};
 
 mod command;
-pub use command::Command;
+pub use command::{Command, OverSsh};
+
+pub(crate) mod escape;
```
NobodyXu, line 172:
```suggestion
mod escape;
```

`lib.rs` is the top-level of a crate, any definition here we be available to all submodules.

src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+    escaped.push(b'\'');
+    OsString::from_vec(escaped).into()
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn test_escape_case(input: &str, expected: &str) {
```
NobodyXu, line 61: It would be great if we can also test some non-utf8 characters, but I don't consider it necessary for this PR to be merged.

src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+/// 
+/// **Note**: This function is an adaptation of [`shell-escape::unix::escape`].
+/// This function exists only for type compatibility and the implementation is
+/// almost exactly the same as [`shell-escape::unix::escape`].
+/// 
+/// [`shell-escape::unix::escape`]: https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#101
+/// 
+pub fn escape(s: &OsStr) -> Cow<'_, OsStr> {
+    let s = s.as_bytes();
+    let all_whitelisted = s.iter().all(|x| whitelisted(*x));
```
NobodyXu, line 32:
```suggestion
    let all_whitelisted = s.iter().copied().all(whitelisted);
```

src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+        return OsString::from_vec(s.to_vec()).into();
+    }
+
+    let mut escaped = Vec::with_capacity(s.len() + 2);
+    escaped.push(b'\'');
+
+    for &b in s {
+        match b {
+            b'\'' | b'!' => {
+                escaped.push(b'\'');
```
NobodyXu, line 44: Nit: I think we can `escaped.reserve(4)` here to reduce allocation.

src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+/// almost exactly the same as [`shell-escape::unix::escape`].
+/// 
+/// [`shell-escape::unix::escape`]: https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#101
+/// 
+pub fn escape(s: &OsStr) -> Cow<'_, OsStr> {
+    let s = s.as_bytes();
+    let all_whitelisted = s.iter().all(|x| whitelisted(*x));
+
+    if !s.is_empty() && all_whitelisted {
+        return OsString::from_vec(s.to_vec()).into();
```
NobodyXu, line 35: You can use `Cow::Borrowed` on the original `s: &OsStr` here.

NobodyXu (changes requested):

NobodyXu:
> okay how about these changes? (thank you for being patient.)

It's much better, but we would still need @jonhoo 's approval on this.
I'm not entirely sure about escaping since the utf encoding is a mess, so I just want to be safe and ask @jonhoo to also review this.

NobodyXu:
@aalekhpatel07 BTW, don't forget to create a new issue on rust-lang/rust and linked it to this.
We would also have to waiting for the compiler guys to investigate and fix the issues.

NobodyXu:
@aalekhpatel07 Hmmm, let me create the bug report for this.

NobodyXu:
@aalekhpatel07 I've created an ICE bug report on rust-lang/rust.

Note that the lint CI did work as intended and you need to fix all `cargo-clippy` warnings.

src/escape.rs
```diff
@@ -0,0 +1,93 @@
...
+pub(crate) fn escape(s: &OsStr) -> Cow<'_, OsStr> {
+    let as_bytes = s.as_bytes();
+    let all_whitelisted = as_bytes.iter().copied().all(whitelisted);
+
+    if !as_bytes.is_empty() && all_whitelisted {
+        return Cow::Borrowed(s);
+    }
+
+    let mut escaped = Vec::with_capacity(as_bytes.len() + 2);
+    escaped.reserve(4);
```
NobodyXu, line 35: Ehh, I don't mean to reserve it here.

```suggestion
```

src/escape.rs
```diff
@@ -0,0 +1,93 @@
...
+    escaped.push(b'\'');
+
+    for &b in as_bytes {
+        match b {
+            b'\'' | b'!' => {
+                escaped.push(b'\'');
```
NobodyXu, line 41: I'd want you to do the reserve here in the match, not outside the loop:

```suggestion
                escaped.reserve(4);
                escaped.push(b'\'');
```

src/escape.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn test_escape_case(input: &str, expected: &str) {
+        let input_os_str = OsStr::from_bytes(input.as_bytes());
+        let observed_os_str = escape(input_os_str);
+        let expected_os_str = OsStr::from_bytes(expected.as_bytes());
+        assert_eq!(observed_os_str, expected_os_str);
```
NobodyXu, line 61:
```suggestion
        test_escape_from_bytes(input.as_bytes(), expected.as_bytes())
```

NobodyXu (changes requested):
Almost done from me, some small changes requested...

Regarding the ICEs, there's already a PR in rust-lang/rust to fix this.

NobodyXu (approved):
Now we just need to wait for @jonhoo to review this and upstream to fix the ICEs on stable channel.

aalekhpatel07:
They have a PR for that ICE already??

Wow! I'm blown away by how responsive, knowledgeable, and supportive the contributors are in the Rust ecosystem. 

Thank you for all your help, @NobodyXu, for helping me improve the PR and develop my understanding of `Cow`s and `OsStr`s in Rust.

NobodyXu:
> They have a PR for that ICE already??

Yes, https://github.com/rust-lang/rust/pull/107532

Though I have no idea on whether it will land in 1.68 or will they have a backport it to 1.67.1 since whether a backport will happen depends on the severity of the bug.

> Thank you for all your help, @NobodyXu, for helping me improve the PR and develop my understanding of `Cow`s and `OsStr`s in Rust.

No worries!

src/command.rs
```diff
@@ -49,6 +51,95 @@ macro_rules! delegate {
...
+    ///         .over_session(&session)
+    ///         .output()
+    ///         .await?;
+    ///
+    ///     assert!(String::from_utf8(ls.stdout).unwrap().contains("total"));
+    /// #   Ok(())
+    /// }
+    ///
+    /// ```
+    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
```
jonhoo, line 87: It's generally common in Rust to have the method of a single-purpose trait match the name of that trait. See for example `Clone::clone`, `AsRef::as_ref`, `Hash::hash`, etc. So we should either name this trait `OverSession` or name the method `over_ssh`. I _think_ I prefer the latter, because it's clearer what's going on from reading just the code. We _could_ also make it `OverSsh::over_ssh_session`, though the verbosity is a little unfortunate.

src/command.rs
```diff
@@ -49,6 +51,95 @@ macro_rules! delegate {
...
 }
 
+/// If a command is `OverSsh` then it can be executed over an SSH session.
+/// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
+pub trait OverSsh {
+    /// Given a session, return a command that can be executed over that session.
+    ///
+    /// **Note**: The command to be executed on the remote machine does not include
+    /// any environment variables or the current directory. They must be
+    /// set explicitly, if desired.
```
jonhoo, line 61: What does "set explicitly" mean here?

jonhoo: Also, I wonder if there's an argument for us simply returning an error if we find that the current working directory or an environment variable is overridden?

Alternatively, can we transparently handle that as well? I guess probably not without assuming the remote end is Bash?

src/command.rs
```diff
@@ -49,6 +51,95 @@ macro_rules! delegate {
     }};
 }
 
+/// If a command is `OverSsh` then it can be executed over an SSH session.
+/// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
```
jonhoo, line 55:
```suggestion
/// If a command is `OverSsh` then it can be executed over an SSH session.
///
/// Primarily a way to allow `std::process::Command` to be turned directly into an `openssh::Command`.
```
Entirely a nit: it will make the rustdoc render a little nicer.

src/command.rs
```diff
@@ -49,6 +51,95 @@ macro_rules! delegate {
...
+    fn over_session<'session>(&self, session: &'session Session) -> crate::Command<'session>;
+}
+
+impl OverSsh for std::process::Command {
+    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
+        let program_escaped: Cow<'_, OsStr> = escape(self.get_program());
+        let mut command = session.raw_command(program_escaped);
+
+        let args = self.get_args().map(escape);
+        command.raw_args(args);
```
jonhoo, line 96: Why use `escape` + `raw_args` rather than just calling `args` directly and have that do the escaping for us?

jonhoo: Same above for `escape` + `raw_command`

src/command.rs
```diff
@@ -49,6 +51,95 @@ macro_rules! delegate {
...
+
+impl<S> OverSsh for std::sync::Arc<S>
+where
+    S: OverSsh,
+{
+    fn over_session<'session>(&self, session: &'session Session) -> Command<'session> {
+        <S as OverSsh>::over_session(self, session)
+    }
+}
+
```
jonhoo, line 142: I think these probably aren't worth including. Even the impl for `Box` feels a little odd. I also think `impl for &mut` is missing, if anything (but again, I think just having the `impl for &` is probably good enough).

src/escape.rs
```diff
@@ -0,0 +1,90 @@
+//! Escape characters that may have special meaning in a shell, including spaces.
+//! This is a modified version of the [`shell-escape::unix`] module of [`shell-escape`] crate.
```
jonhoo, line 2: Can we submit a PR to `shell-escape` to add a version supporting `OsStr` there directly? I'm okay with us landing our own version initially, but it'd be way better if we could just use a version from them in the future.

src/escape.rs
```diff
@@ -0,0 +1,90 @@
...
+//! [`shell-escape::unix`]: https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#101
+
+use std::{
+    borrow::Cow,
+    ffi::{OsStr, OsString},
+    os::unix::ffi::OsStrExt,
+    os::unix::ffi::OsStringExt,
+};
+
+fn whitelisted(byte: u8) -> bool {
```
jonhoo, line 14: Super nit, but let's call this "allowed".

tests/openssh.rs
```diff
@@ -292,6 +292,31 @@ async fn stdout() {
     }
 }
 
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn over_session() {
```
jonhoo, line 297: I'd like to see this also test some cases that require escaping.

jonhoo (changes requested):
This is a really neat idea! Thank you both for pushing it forward. I've left some notes.

src/command.rs
NobodyXu: Because command/args only takes str and here we have OsStr.

Once this PR is merged, I plan to use escape introduced here for OsStr inside command/args so that they can take OsStr.

src/command.rs
NobodyXu: I think impl for &T can also handle &mut T given rust auto-deref rules?

src/command.rs
NobodyXu: The mux protocol actually supports env specification, though I didn't expose that because openssh didn't have that method.

For setting current dir, AFAIK, ssh does not support that.

jonhoo: In that case, we should error (if possible) if `Command::current_dir` or `Command::env` has been called on the `Command` we're passed in.

src/command.rs
jonhoo: I think that's only sort of true. In particular it still means that `&mut OverSssh` doesn't implement `OpenSsh`, which can be annoying in certain cases.

src/escape.rs
aalekhpatel07: Great suggestion! I've initiated [a PR](https://github.com/sfackler/shell-escape/pull/9) to `shell-escape` and if it gets merged, we can avoid having a `src/escape.rs` altogether.

src/command.rs
aalekhpatel07: I do believe `OverSsh::over_ssh` makes it quite a bit clearer. That's an excellent improvement. Thank you.

tests/openssh.rs
aalekhpatel07: I'm not quite sure what a good test case should look like here and I'd greatly appreciate a hint. :thinking:

aalekhpatel07:
one more set of changes as suggested in the review. please lmk if I missed something or misinterpreted a suggestion.

tests/openssh.rs
jonhoo: Ah, so, I was thinking things like having an argument that contains, say, `'`, ` `, and `\`.

src/command.rs
```diff
@@ -49,6 +51,129 @@ macro_rules! delegate {
...
+    /// ```no_run
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    ///     use std::process::Command;
+    ///     use openssh::{Session, KnownHosts, OverSsh};
+    ///
+    ///     let session = Session::connect_mux("me@ssh.example.com", KnownHosts::Strict).await?;
+    ///     let echo =
+    ///         Command::new("echo")
+    ///         .arg("$MY_ENV_VAR")
```
jonhoo, line 106: This doesn't seem right — did you mean `.env`?

src/command.rs
```diff
@@ -49,6 +51,129 @@ macro_rules! delegate {
...
+    ///         .over_ssh(&session);
+    ///     assert_matches!(echo, Err(openssh::Error::CommandHasEnv));
```
jonhoo, line 108: This doesn't currently compile.

jonhoo (changes requested):

src/command.rs
aalekhpatel07: You're absolutely right. That is an oversight on my part. Should've been `Command::new("echo").env("MY_ENV_VAR", "foo")`

src/command.rs
aalekhpatel07: hmm. after the `.env` fix, maybe we should use `assert!(matches!(...))` instead.

tests/openssh.rs
```diff
@@ -292,6 +292,83 @@ async fn stdout() {
...
+    }
+}
+
+
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn over_session_ok_require_escaping_arguments() {
+    for session in connects().await {
+        let mut command = std::process::Command::new("echo")
+            .arg("\"\'\'foo\'\'\"")
```
jonhoo, line 326: I'd like to see a space (` `) in here too :)

jonhoo (approved):
Thanks for sticking with this!

jonhoo:
Looks like rustfmt is complaining, and somehow we're also hitting a compiler ICE?
https://github.com/openssh-rust/openssh/actions/runs/4153289081/jobs/7184837200

aalekhpatel07:
Really appreciate yours and @NobodyXu's  precious time reviewing this! I was able to learn a few things from this review, so thank you! Now hopefully the other PR in `shell-escape` also lands soon.

aalekhpatel07:
yep the compiler ICE was pointed out by @NobodyXu last week and I believe there was some work happening in rust-lang/rust#107532 to fix it. Rustfmt should be happy now though.

jonhoo:
I'll hold off on merging this until the fix for the ICE lands then 👍 Nothing more needed from your end at this point :)

NobodyXu:
Since the ICE is fixed and the CI has passed, I will merge this PR.

PR #120, opened 2023-06-02
Don't attempt to resolve TCP sockets client side

benesch:
It may be important to have the SSH server perform DNS resolution, as the client often does not use the same DNS server as the server.

We ran into this issue at @MaterializeInc, where we use `openssh` to set up SSH tunnels via bastion hosts. Our customers often want to establish a tunnel to connect to a host whose name is not resolvable on the client side of the tunnel.

### DIFF
--- a/src/port_forwarding.rs
+++ b/src/port_forwarding.rs
@@ -6,8 +6,7 @@ use std::ffi::OsStr;
 
 use std::borrow::Cow;
 use std::fmt;
-use std::io;
-use std::net::{self, SocketAddr, ToSocketAddrs};
+use std::net::{self, SocketAddr};
 use std::path::{Path, PathBuf};
 
 /// Type of forwarding
@@ -44,12 +43,20 @@ pub enum Socket<'a> {
     },
 
     /// Tcp socket.
-    TcpSocket(SocketAddr),
+    TcpSocket {
+        /// Hostname.
+        host: Cow<'a, str>,
+        /// Port.
+        port: u16,
+    },
 }
 
 impl From<SocketAddr> for Socket<'static> {
     fn from(addr: SocketAddr) -> Self {
-        Socket::TcpSocket(addr)
+        Socket::TcpSocket {
+            host: addr.ip().to_string().into(),
+            port: addr.port(),
+        }
     }
 }
 
@@ -99,11 +106,14 @@ impl From<Box<Path>> for Socket<'static> {
 
 impl Socket<'_> {
     /// Create a new TcpSocket
-    pub fn new<T: ToSocketAddrs>(addr: &T) -> Result<Socket<'static>, io::Error> {
-        addr.to_socket_addrs()?
-            .next()
-            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "no more socket addresses to try"))
-            .map(Socket::TcpSocket)
+    pub fn new<'a, S>(host: S, port: u16) -> Socket<'a>
+    where
+        S: Into<Cow<'a, str>>,
+    {
+        Socket::TcpSocket {
+            host: host.into(),
+            port,
+        }
     }
 
     #[cfg(feature = "process-mux")]
@@ -111,7 +121,7 @@ impl Socket<'_> {
         match self {
             #[cfg(unix)]
             Socket::UnixSocket { path } => Cow::Borrowed(path.as_os_str()),
-            Socket::TcpSocket(socket) => Cow::Owned(format!("{}", socket).into()),
+            Socket::TcpSocket { host, port } => Cow::Owned(format!("{host}:{port}").into()),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/120)

src/port_forwarding.rs
```diff
@@ -99,19 +106,19 @@ impl From<Box<Path>> for Socket<'static> {
 
 impl Socket<'_> {
     /// Create a new TcpSocket
```
NobodyXu, line 108: We can keep this function, having a function that accepts ToSocketAddrw can be useful.

src/port_forwarding.rs
```diff
@@ -139,7 +146,7 @@ impl<'a> fmt::Display for Socket<'a> {
             Socket::UnixSocket { path } => {
                 write!(f, "{}", path.to_string_lossy())
```
NobodyXu, line 147:
```suggestion
                write!(f, "{}", path.display())
```

NobodyXu (changes requested):

src/port_forwarding.rs
benesch: Done.

src/port_forwarding.rs
benesch: My possibly controversial take is that making it easy to call `ToSocketAddrs` on the client is likely to lead to more bugs like the one that led me to file this PR! When using SSH port forwarding to tunnel through an SSH bastion host—which I imagine is the common use case—it's extremely likely that the client cannot resolve the hostname, because the hostname is private to the network with the bastion host. ISTM like the API here should encourage deferring resolution of hostnames, so that `openssh` can perform the resolution on the server if appropriate.

benesch:
Thanks for the review, @NobodyXu. Sorry for the terse PR description last night; filed in a rush. Added some more context to the description now.

NobodyXu:
@benesch Please run `cargo fmt` locally and fix the format.
Thanks!

src/port_forwarding.rs
NobodyXu: Got it, this sounds reasonable to me.

benesch:
Whoops, sorry about that! Done.

NobodyXu (approved):

src/port_forwarding.rs
```diff
@@ -99,19 +106,22 @@ impl From<Box<Path>> for Socket<'static> {
...
 impl Socket<'_> {
     /// Create a new TcpSocket
-    pub fn new<T: ToSocketAddrs>(addr: &T) -> Result<Socket<'static>, io::Error> {
-        addr.to_socket_addrs()?
-            .next()
-            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "no more socket addresses to try"))
-            .map(Socket::TcpSocket)
+    pub fn new<'a, S>(host: S, port: u16) -> Socket<'a>
+    where
+        S: Into<Cow<'a, str>>,
```
jonhoo, line 111: This changes the public api, and so would end up requiring a major version bump. Would you mind making this a separate constructor instead?

cc @NobodyXu we have to be a little careful about these kinds of changes. Maybe we should add a CI step that invokes https://github.com/obi1kenobi/cargo-semver-checks or https://github.com/Enselic/cargo-public-api to catch these things automatically (at least in most cases).

NobodyXu: > This changes the public api, and so would end up requiring a major version bump. Would you mind making this a separate constructor instead?

Sure, but with #131 we probably need a breaking change anyway.

> cc @NobodyXu we have to be a little careful about these kinds of changes. Maybe we should add a CI step that invokes https://github.com/obi1kenobi/cargo-semver-checks or https://github.com/Enselic/cargo-public-api to catch these things automatically (at least in most cases).

Yeah, we definitely need semver-checking in CI.

stephan-cr: I'm sorry to chime in, I saw the stream and think there's common mistake regarding SemVer and the zero major version you might not be aware of (and I learned recently as well). :-)

My comment is just for info, choose whatever version you want. I just want to give some hint.

> This changes the public api, and so would end up requiring a major version bump.

Technically, there's no need to bump the major version, because 1.0.0 hasn't been released yet. Therefore, everything so far is "initial development".

The following is cited from [SemVer spec](https://semver.org/#spec-item-4):

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

NobodyXu: Thanks!

jonhoo: @stephan-ci in Rust, when the leftmost version number component is a 0, then the next digit over is effectively considered the major version. That's what I was referring to bumping. Updating x in 0.x has the same effect and implications as bumping x in x.0, and so it is still considered (and referred to) as a major version bump.

benesch: Just a user here, so discount my opinion appropriately, but strong preference from me for a semver-breaking release with _just_ the new API, rather than a backcompatible release that includes both APIs! The existing API is, AFAICT, not possible to use correctly with remote port forwarding, and so forcing users into the new API seems worth the breakage to me.

(Earlier discussion on this: https://github.com/openssh-rust/openssh/pull/120#discussion_r1215492541)

PR #126, opened 2023-07-28
support clean up the historical control_directory 

### LINKED ISSUE #125: Is it possible to clean up the historical control_directory before creating a SessionBuilder?
Currently testing locally, the historical [control_directory](https://docs.rs/openssh/latest/openssh/struct.SessionBuilder.html#method.control_directory) directory will always exist during each debugging session.

E.g:
```shell
➜ ll -als
total 0
0 drwxr-xr-x  6 root  staff   192B  7 28 10:03 .
0 drwxr-xr-x  3 root  staff    96B  7 27 18:25 ..
0 drwxr-xr-x  4 root  staff   128B  7 28 09:55 .ssh-connectionpCn9J7
0 drwxr-xr-x  2 root  staff    64B  7 28 10:01 .ssh-connectionpHq9a2
0 drwxr-xr-x  2 root  staff    64B  7 28 10:03 .ssh-connectionpJp983
0 drwxr-xr-x  2 root  staff    64B  7 28 10:05 .ssh-connectionpLwp24
```

baoyachi:
fix https://github.com/openssh-rust/openssh/issues/125

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -164,6 +164,9 @@ impl SessionBuilder {
     /// in directory specified by [`SessionBuilder::control_directory`], created by
     /// previous `openssh::Session` that is not cleaned up for some reasons
     /// (e.g. process getting killed, abort on panic, etc)
+    ///
+    /// Use this with caution, do not enable this if you don't understand
+    /// what it does,
     #[cfg(not(windows))]
     #[cfg_attr(docsrs, doc(cfg(not(windows))))]
     pub fn clean_history_control_directory(&mut self, clean: bool) -> &mut Self {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -34,28 +34,23 @@ fn get_default_control_dir<'a>() -> Result<&'a Path, Error> {
         })
 }
 
-fn clean_history_control_dir(dir: &TempDir, prefix: &str) -> io::Result<()> {
-    // Check if the parent directory of the given TempDir exists
-    if let Some(parent) = dir.path().parent() {
-        // Read the entries in the parent directory
-        fs::read_dir(parent)?
-            // Filter out and keep only the valid entries
-            .filter_map(Result::ok)
-            // Filter the entries to only include files that start with prefix
-            .filter(|entry| {
-                if let Ok(file_type) = entry.file_type() {
-                    file_type.is_dir()
-                        && entry.file_name().to_string_lossy().starts_with(prefix)
-                        && entry.path() != dir.path()
-                } else {
-                    false
-                }
-            })
-            // For each matching entry, remove the directory
-            .for_each(|entry| {
-                let _ = fs::remove_dir_all(entry.path());
-            });
-    }
+fn clean_history_control_dir(socketdir: &Path, prefix: &str) -> io::Result<()> {
+    // Read the entries in the parent directory
+    fs::read_dir(socketdir)?
+        // Filter out and keep only the valid entries
+        .filter_map(Result::ok)
+        // Filter the entries to only include files that start with prefix
+        .filter(|entry| {
+            if let Ok(file_type) = entry.file_type() {
+                file_type.is_dir() && entry.file_name().to_string_lossy().starts_with(prefix)
+            } else {
+                false
+            }
+        })
+        // For each matching entry, remove the directory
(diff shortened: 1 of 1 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/126)

src/builder.rs
```diff
@@ -34,6 +34,31 @@ fn get_default_control_dir<'a>() -> Result<&'a Path, Error> {
         })
 }
 
+fn clean_history_control_dir(dir: &TempDir, prefix: &str) -> io::Result<()> {
+    // Check if the parent directory of the given TempDir exists
+    if let Some(parent) = dir.path().parent() {
+        // Read the entries in the parent directory
+        fs::read_dir(parent)?
+            // Filter out and keep only the valid entries
+            .filter_map(|entry| entry.ok())
```
NobodyXu, line 43:
```suggestion
            .filter_map(Result::ok)
```

src/builder.rs
```diff
@@ -133,6 +160,24 @@ impl SessionBuilder {
...
+    ///
+    /// Generally, the configuration of 'control_directory' is usually a Temporary folder.
+    /// Please refer to issue:https://github.com/openssh-rust/openssh/pull/28
+    //
+    // If not configured, openssh will automatically create a temporary .ssh_connection dir in the
+    // current running environment.
+    //
+    // This method provides control for cleaning up history 'control_directory' function,
+    // default openssh will not clean up. If necessary, you can manually call this method
+    ///
```
NobodyXu, line 173:
```suggestion
    /// Clean up the temporary directories with the `.ssh-connection` prefix
    /// in directory specified by [`SessionBuilder::control_directory`], created by
    /// previous `openssh::Session` that is not cleaned up for some reasons
    /// (e.g. process getting killed, abort on panic, etc)
```

NobodyXu (changes requested):

src/builder.rs
```diff
@@ -133,6 +160,17 @@ impl SessionBuilder {
         self
     }
 
+    /// Clean up the temporary directories with the `.ssh-connection` prefix
+    /// in directory specified by [`SessionBuilder::control_directory`], created by
+    /// previous `openssh::Session` that is not cleaned up for some reasons
+    /// (e.g. process getting killed, abort on panic, etc)
```
NobodyXu, line 166:
```suggestion
    /// (e.g. process getting killed, abort on panic, etc)
    ///
    /// Use this with caution, do not enable this if you don't understand
    /// what it does,
```

NobodyXu (changes requested):

src/builder.rs
```diff
@@ -295,11 +336,16 @@ impl SessionBuilder {
...
+        let prefix = ".ssh-connection";
         let dir = Builder::new()
-            .prefix(".ssh-connection")
+            .prefix(prefix)
             .tempdir_in(socketdir)
             .map_err(Error::Master)?;
 
+        if self.clean_history_control_dir {
+            let _ = clean_history_control_dir(&dir, prefix);
+        }
```
NobodyXu, line 347: You can simplify this by calling `clean_history_control_dir` before creating the new `TempDir`, then you won't have to compare the `entry.path() != dir.path()` and you would also don't need to call `dir.path().parent()` with a `if-let`

NobodyXu (changes requested):

src/builder.rs
baoyachi: That's a good suggestion.

NobodyXu:
P.S. I would recommend rebasing against main instead of merging main into this PR, which will create an additional commit.

NobodyXu (approved):

jonhoo:
I assume these mostly stuck around for when a connection failed, which has since been fixed (I think). I'm not in love with doing a recursive directory remove here invisibly on the user's behalf — feels like a recipe for accidentally deleting something unintentional down the line.

NobodyXu:
> I assume these mostly stuck around for when a connection failed, which has since been fixed (I think). I'm not in love with doing a recursive directory remove here invisibly on the user's behalf — feels like a recipe for accidentally deleting something unintentional down the line.

This behavior is disabled by default, so I think it might be fine.

PR #128, opened 2023-07-31
Public method for creating a Session

### LINKED ISSUE #124: Make Session function public for external calls
Currently, I have implemented a Session object myself. Can these two methods be opened for external self control calls.

Make `new_process_mux` or `new_native_mux` public .

```diff

mpl Session {
    #[cfg(feature = "process-mux")]
+    pub(super) fn new_process_mux(tempdir: TempDir) -> Self {
        Self(SessionImp::ProcessImpl(process_impl::Session::new(tempdir)))
    }

    #[cfg(feature = "native-mux")]
+    pub(super) fn new_native_mux(tempdir: TempDir) -> Self {
        Self(SessionImp::NativeMuxImpl(native_mux_impl::Session::new(
            tempdir,
        )))
    }
}
```

[About code link ref:](https://github.com/openssh-rust/openssh/blob/3319a8162a/src/session.rs#L56-L65)

baoyachi:
fix #124

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -302,7 +302,6 @@ impl SessionBuilder {
     /// # Examples
     ///
     /// ```rust
-    /// use std::borrow::Cow;
     /// use openssh::SessionBuilder;
     /// let b = SessionBuilder::default();
     /// let (b, d) = b.resolve("ssh://test-user@127.0.0.1:2222");
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -348,7 +348,8 @@ impl SessionBuilder {
         (Cow::Owned(with_overrides), destination)
     }
 
-    /// Create ssh master session and return [`TempDir`]
+    /// Create ssh master session and return [`TempDir`] which
+    /// contains the ssh control socket.
     pub async fn launch_master(&self, destination: &str) -> Result<TempDir, Error> {
         let socketdir = if let Some(socketdir) = self.control_dir.as_ref() {
             socketdir
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -93,6 +93,16 @@ impl Default for SessionBuilder {
 }
 
 impl SessionBuilder {
+    /// Return the user set in builder.
+    pub fn get_user(&self) -> Option<&str> {
+        self.user.as_deref()
+    }
+
+    /// Return the port set in builder.
+    pub fn get_port(&self) -> Option<&str> {
+        self.port.as_deref()
+    }
+
     /// Set the ssh user (`ssh -l`).
     ///
     /// Defaults to `None`.
@@ -286,7 +296,21 @@ impl SessionBuilder {
         Ok(f(tempdir))
     }
 
-    fn resolve<'a, 'b>(&'a self, mut destination: &'b str) -> (Cow<'a, Self>, &'b str) {
+    /// [`SessionBuilder`] support for `destination` parsing.
+    /// The format of `destination` is the same as the `destination` argument to `ssh`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use std::borrow::Cow;
+    /// use openssh::SessionBuilder;
+    /// let b = SessionBuilder::default();
+    /// let (b, d) = b.resolve("ssh://test-user@127.0.0.1:2222");
+    /// assert_eq!(b.get_port().as_deref(), Some("2222"));
+    /// assert_eq!(b.get_user().as_deref(), Some("test-user"));
+    /// assert_eq!(d, "127.0.0.1");
+    /// ```
+    pub fn resolve<'a, 'b>(&'a self, mut destination: &'b str) -> (Cow<'a, Self>, &'b str) {
         // the "new" ssh://user@host:port form is not supported by all versions of ssh,
         // so we always translate it into the option form.
         let mut user = None;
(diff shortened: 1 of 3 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/128)

src/session.rs
```diff
@@ -53,13 +53,107 @@ pub struct Session(SessionImp);
...
+    /// #[derive(Clone, Debug, Default)]
+    /// pub struct CustomSessionBuilder {
+    ///    user: Option<String>,
+    ///    port: Option<String>,
+    ///    password: Option<String>,
+    /// }
+    ///
+    /// let builder = CustomSessionBuilder::default().password("123456");
+    /// let (builder, destination) = builder.resolve("ssh://jon@ssh.thesquareplanet.com:222");
+    /// let tempdir = builder.launch_master(destination).await?;
```
NobodyXu, line 81: I would rather the examples to be much simpler and pass the compilation.

src/session.rs
```diff
@@ -53,13 +53,107 @@ pub struct Session(SessionImp);
...
 
 impl Session {
+    /// The method for creating a [`Session`] and externally control the creation of TempDir.
+    ///
+    /// By using the built-in [`SessionBuilder`] in openssh, or a custom SessionBuilder,
+    /// create a TempDir.
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
```
NobodyXu, line 63: I'm ok with `no_run` but not ok with ignoring the example altogether:

```suggestion
    /// ```no_run
```

NobodyXu (changes requested):

src/session.rs
baoyachi: That's a good idea.

src/session.rs
baoyachi: I will write an example to modify it

src/builder.rs
```diff
@@ -286,7 +286,7 @@ impl SessionBuilder {
         Ok(f(tempdir))
     }
 
-    fn resolve<'a, 'b>(&'a self, mut destination: &'b str) -> (Cow<'a, Self>, &'b str) {
+    pub fn resolve<'a, 'b>(&'a self, mut destination: &'b str) -> (Cow<'a, Self>, &'b str) {
```
NobodyXu, line 289: You need to provide doc for this public method.

NobodyXu (changes requested):

src/builder.rs
```diff
@@ -324,7 +324,7 @@ impl SessionBuilder {
         (Cow::Owned(with_overrides), destination)
     }
 
-    async fn launch_master(&self, destination: &str) -> Result<TempDir, Error> {
+    pub async fn launch_master(&self, destination: &str) -> Result<TempDir, Error> {
```
NobodyXu, line 327: Same for this, doc is required.

NobodyXu (changes requested):

src/builder.rs
```diff
@@ -286,7 +286,22 @@ impl SessionBuilder {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use std::borrow::Cow;
+    /// use openssh::SessionBuilder;
+    /// let b = SessionBuilder::default();
+    /// let (b, d) = b.resolve("ssh://test-user@127.0.0.1:2222");
+    /// assert_eq!(d, "127.0.0.1");
+    /// assert_eq!(b, Cow::Borrowed(SessionBuilder::default()
```
NobodyXu, line 300: You could instead, add a getter for retrieving `port` and `user` of `SessioBuilder`.

NobodyXu, line 300: You could instead, add a getter for retrieving `port` and `user` of `SessioBuilder`.

src/builder.rs
```diff
@@ -55,7 +55,7 @@ fn clean_history_control_dir(socketdir: &Path, prefix: &str) -> io::Result<()> {
 }
 
 /// Build a [`Session`] with options.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
NobodyXu, line 58: I would not like `PartialEq` to be added to `SessionBuilder`, it should be an opaque struct and we do not provide any guarantee on how it is stored or the order of them.

NobodyXu (changes requested):

src/builder.rs
baoyachi: port or use not pub use in doc

NobodyXu: I mean adding a getter for that:

```
/// Return the user set in builder.
pub fn get_user(&self) -> Option<&str> {
    self.user.as_deref()
}

Comment (NobodyXu) [reply src/builder.rs:300]: I mean adding a getter for that:

```
/// Return the user set in builder.
pub fn get_user(&self) -> Option<&str> {
    self.user.as_deref()
}

src/builder.rs
```diff
@@ -286,7 +296,21 @@ impl SessionBuilder {
...
     }
 
-    fn resolve<'a, 'b>(&'a self, mut destination: &'b str) -> (Cow<'a, Self>, &'b str) {
+    /// [`SessionBuilder`] support for `destination` parsing.
+    /// The format of `destination` is the same as the `destination` argument to `ssh`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use std::borrow::Cow;
```
NobodyXu, line 305:
```suggestion
```

src/builder.rs
```diff
@@ -324,7 +348,8 @@ impl SessionBuilder {
         (Cow::Owned(with_overrides), destination)
     }
 
-    async fn launch_master(&self, destination: &str) -> Result<TempDir, Error> {
+    /// Create ssh master session and return [`TempDir`]
```
NobodyXu, line 351:
```suggestion
    /// Create ssh master session and return [`TempDir`] which
    /// contains the ssh control socket.
```

NobodyXu (changes requested):

src/builder.rs
baoyachi: @NobodyXu  .How to find quick this code warning ?

NobodyXu (approved):

Issue #119, opened 2023-05-25
how to use Username and password authetication


baoyachi:
Is there have an example of connect on SFTP username and password?

NobodyXu:
Transfer is done.

For openssh, the only way to use password authentication is to create the ssh multiplex master yourself and asks password from the user, then use [`Session::resume`](https://docs.rs/openssh/latest/openssh/struct.Session.html#method.resume) or [`Session::resume_mux`](https://docs.rs/openssh/latest/openssh/struct.Session.html#method.resume_mux) to create a `openssh::Session`.

Then you can spawn sftp session by simply calling [`Sftp::from_session`](https://docs.rs/openssh-sftp-client/latest/openssh_sftp_client/struct.Sftp.html#method.from_session) on the `Session` you get.

baoyachi:
Thx,I'm try it

baoyachi:
Casued error: "kex_exchange_identification: connection closed by remote host"

NobodyXu:
@baoyachi Can you show me the code so that I can help you debug the problem?

baoyachi:
I will send PR later to solve this issue.

baoyachi:
@NobodyXu Currently, openssh supports username, and the password can be [seen here](https://github.com/baoyachi/openssh-extension/blob/main/src/openssh_ext.rs#L313-L335). If possible, I can submit a PR.

---
[openssh-extension](https://github.com/baoyachi/openssh-extension): Based on [openssh](https://github.com/openssh-rust/openssh) extension。

Currently, [openssh](https://github.com/openssh-rust/openssh) only supports [public key authentication](https://github.com/openssh-rust/openssh/blob/v0.9.9/src/session.rs#L103-L105) methods. However, in certain specific scenarios, it may still be necessary to use user password for sftp login.

This repository is based on Openssh and extends support for sftp user password authentication.

NobodyXu:
@baoyachi I'm willing to accept a PR for this.

Regarding the opendal, isn't there already official support for sftp using openssh + openssh_sftp_client?

baoyachi:
Yes, currently opendal does not support SFTP username and password login. Due to different dependencies, I have implemented compatibility. If this PR is merged by openssh, openssh-sftp-client,the latest version of openssh, openssh-sftp-client feature can be updated in the future of opendal

NobodyXu:
Thanks for explanation, I am willing to accpet PR for password login.

NobodyXu:
[...]

PR #129, opened 2023-08-03
support user password login ssh

baoyachi:
fixed: #119

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -302,7 +302,6 @@ impl SessionBuilder {
     /// # Examples
     ///
     /// ```rust
-    /// use std::borrow::Cow;
     /// use openssh::SessionBuilder;
     /// let b = SessionBuilder::default();
     /// let (b, d) = b.resolve("ssh://test-user@127.0.0.1:2222");
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -348,7 +348,8 @@ impl SessionBuilder {
         (Cow::Owned(with_overrides), destination)
     }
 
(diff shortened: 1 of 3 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/129)

src/builder.rs
```diff
@@ -60,6 +60,7 @@ pub struct SessionBuilder {
     user: Option<String>,
     port: Option<String>,
     keyfile: Option<PathBuf>,
+    password: Option<String>,
```
NobodyXu, line 63: IMHO we can use [zeroize](https://docs.rs/zeroize/) here to avoid leaking the password.

src/session.rs
```diff
@@ -194,6 +195,35 @@ impl Session {
...
+    /// password).
+    ///
+    /// Warning ⚠️: current use sshpass non-interactive ssh password auth.
+    /// See detail: <https://sourceforge.net/projects/sshpass/>
+    ///
+    /// The current implementation of sshpass is only temporary, and the functionality will be
+    /// optimized later. If using this feature, pay attention to security risks.
+    ///
+    /// For more options, see [`SessionBuilder`].
+    #[cfg(feature = "process-mux")]
```
NobodyXu, line 214: I'd like to put it under another feature "ssh-pass-login" since in the future we will implement this ourselves using pty, which will pull in additional features most don't use.

NobodyXu (changes requested):

NobodyXu:
@baoyachi didn't you accidentally delete the repository?

PR #136, opened 2023-09-16
Attempt at providing an owned version of command

### LINKED ISSUE #117: Feature: A new `RemoteChildOwned` struct to avoid self-ref issue
From https://github.com/openssh-rust/openssh-sftp-client/pull/65#issuecomment-1533280291 :

> If it possible for Sftp to hold a session and spawn command (which means hold on RemoteChild) internally? So that users don't need to hold session, remote_child and Sftp at the same time. To avoid RemoteChild been dropped earlier than Sftp, users like opendal have to use owning_ref tricks.

`openssh_sftp_client::Sftp` now has a new method `new_with_auxiliary` to hold any data and @Xuanwo & @silver-ymz is currently using a [workaround I proposed](https://github.com/openssh-rust/openssh-sftp-client/pull/65#issuecomment-1535137285) by using a `Box`ed future.

But I think it's still better to have a `RemoteChildOwned` to avoid the self-ref issue, it's pretty bad for anybody using it with `openssh_sftp_client`.

Alternatively, we can wait for the new proxy-mode to openssh-mux-client https://github.com/openssh-rust/openssh-mux-client/issues/6 , which will enable you to use stdin/stdout/stderr while dropping Session and RemoteChild.

It will also support windows.

Unfortunately, it was blocking on `tokio_util::sync::WaitForCancellationFutureOwned`, so I switched to other projects and I kind of forgot it.
I will restart working on it, but it will takes quite some time before it's ready to use.

masklinn:
Pushing this to run the tests as getting things to work locally on a mac seems like a bit of a bear.

- Could not get the lifetimes right to not break `RemoteChild::session` (the lifetime becomes bound to the child's rather than the original session's), so move it into its own impl block, and added an extremely debatable `Child::into_session`, since it means not being able to call `wait`, or `wait_with_output`, or `disconnect`, so probably not useful (but a slightly separate variant of `Child::session` seems awful).
- Initially wanted to have a version directly owning a `Session`, but that turns out impossible as the `Command` builder works off of mutable references, so not possible to consume the builder in order to get the session out, short of creating either separate `impl` blocks for `Command<&Session>`, `Command<Arc<Session>>` and `Command<Session>`, or separate consuming spawns, neither sounding great.
- Naming is dubious, especially `shared_command`, but I don't recall ever seeing a method around shared refs so I went simple. I would have impl'd `Arc<Session>::into_command` but that's not legal...

  Maybe a free function would be better?

Hopefully will eventually fix #117

### DIFF
--- a/src/child.rs
+++ b/src/child.rs
@@ -1,4 +1,4 @@
-use super::{ChildStderr, ChildStdin, ChildStdout, Error, Session};
+use super::{ChildStderr, ChildStdin, ChildStdout, Error};
 
 use std::io;
 use std::process::{ExitStatus, Output};
@@ -42,27 +42,30 @@ macro_rules! delegate {
 
 /// Representation of a running or exited remote child process.
 ///
-/// This structure is used to represent and manage remote child processes. A remote child process
-/// is created via the [`Command`](crate::Command) struct through [`Session::command`], which
-/// configures the spawning process and can itself be constructed using a builder-style interface.
+/// This structure is used to represent and manage remote child
+/// processes. A remote child process is created via the
+/// [`OwningCommand`](crate::OwningCommand) struct through
+/// [`Session::command`](crate::Session::command) or one of its
+/// variants, which configures the spawning process and can itself be
+/// constructed using a builder-style interface.
 ///
-/// Calling [`wait`](RemoteChild::wait) (or other functions that wrap around it) will make the
+/// Calling [`wait`](Child::wait) (or other functions that wrap around it) will make the
 /// parent process wait until the child has actually exited before continuing.
 ///
-/// Unlike [`std::process::Child`], `RemoteChild` *does* implement [`Drop`], and will terminate the
+/// Unlike [`std::process::Child`], `Child` *does* implement [`Drop`], and will terminate the
 /// local `ssh` process corresponding to the remote process when it goes out of scope. Note that
 /// this does _not_ terminate the remote process. If you want to do that, you will need to kill it
 /// yourself by executing a remote command like `pkill` to kill it on the remote side.
 ///
-/// As a result, `RemoteChild` cannot expose `stdin`, `stdout`, and `stderr` as fields for
+/// As a result, `Child` cannot expose `stdin`, `stdout`, and `stderr` as fields for
 /// split-borrows like [`std::process::Child`] does. Instead, it exposes
-/// [`stdin`](RemoteChild::stdin), [`stdout`](RemoteChild::stdout),
-/// and [`stderr`](RemoteChild::stderr) as methods. Callers can call `.take()` to get the same
+/// [`stdin`](Child::stdin), [`stdout`](Child::stdout),
+/// and [`stderr`](Child::stderr) as methods. Callers can call `.take()` to get the same
 /// effect as a split borrow and use multiple streams concurrently. Note that for the streams to be
 /// available,`Stdio::piped()` should be passed to the corresponding method on
-/// [`Command`](crate::Command).
+/// [`OwningCommand`](crate::OwningCommand).
 ///
-/// NOTE that once `RemoteChild` is dropped, any data written to `stdin` will not be sent to the
+/// NOTE that once `Child` is dropped, any data written to `stdin` will not be sent to the
 /// remote process and `stdout` and `stderr` will yield EOF immediately.
 ///
 /// ```rust,no_run
@@ -74,8 +77,8 @@ macro_rules! delegate {
 /// # }
 /// ```
 #[derive(Debug)]
-pub struct RemoteChild<'s> {
-    session: &'s Session,
+pub struct Child<S> {
+    session: S,
     imp: RemoteChildImp,
 
     stdin: Option<ChildStdin>,
@@ -83,9 +86,9 @@ pub struct RemoteChild<'s> {
     stderr: Option<ChildStderr>,
 }
 
-impl<'s> RemoteChild<'s> {
+impl<S> Child<S> {
     pub(crate) fn new(
-        session: &'s Session,
+        session: S,
         (imp, stdin, stdout, stderr): (
             RemoteChildImp,
             Option<ChildStdin>,
@@ -100,11 +103,6 @@ impl<'s> RemoteChild<'s> {
             stderr,
             imp,
         }
-    }
-
-    /// Access the SSH session that this remote process was spawned from.
-    pub fn session(&self) -> &'s Session {
-        self.session
     }
 
     /// Disconnect from this given remote child process.
@@ -200,5 +198,12 @@ impl<'s> RemoteChild<'s> {
     /// Access the handle for reading from the remote child's standard error (stderr), if requested.
     pub fn stderr(&mut self) -> &mut Option<ChildStderr> {
         &mut self.stderr
+    }
+}
+
+impl<S: Clone> Child<S> {
+    /// Access the SSH session that this remote process was spawned from.
+    pub fn session(&self) -> S {
+        self.session.clone()
     }
 }
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,12 +1,13 @@
 use crate::escape::escape;
 
+use super::child::Child;
 use super::stdio::TryFromChildIo;
-use super::RemoteChild;
 use super::Stdio;
 use super::{Error, Session};
 
 use std::borrow::Cow;
 use std::ffi::OsStr;
+use std::ops::Deref;
 use std::process;
 
 #[derive(Debug)]
@@ -112,17 +113,17 @@ pub trait OverSsh {
     /// }
     ///
     /// ```
-    fn over_ssh<'session>(
+    fn over_ssh<S: Deref<Target = Session> + Clone>(
         &self,
-        session: &'session Session,
-    ) -> Result<crate::Command<'session>, crate::Error>;
+        session: S,
+    ) -> Result<OwningCommand<S>, crate::Error>;
 }
 
 impl OverSsh for std::process::Command {
-    fn over_ssh<'session>(
+    fn over_ssh<S: Deref<Target = Session> + Clone>(
         &self,
-        session: &'session Session,
-    ) -> Result<Command<'session>, crate::Error> {
+        session: S,
+    ) -> Result<OwningCommand<S>, crate::Error> {
         // I'd really like `!self.get_envs().is_empty()` here, but that's
         // behind a `exact_size_is_empty` feature flag.
         if self.get_envs().len() > 0 {
@@ -134,7 +135,7 @@ impl OverSsh for std::process::Command {
         }
 
         let program_escaped: Cow<'_, OsStr> = escape(self.get_program());
-        let mut command = session.raw_command(program_escaped);
+        let mut command = Session::to_raw_command(session, program_escaped);
 
         let args = self.get_args().map(escape);
         command.raw_args(args);
@@ -143,10 +144,10 @@ impl OverSsh for std::process::Command {
 }
 
 impl OverSsh for tokio::process::Command {
-    fn over_ssh<'session>(
+    fn over_ssh<S: Deref<Target = Session> + Clone>(
         &self,
-        session: &'session Session,
-    ) -> Result<Command<'session>, crate::Error> {
+        session: S,
+    ) -> Result<OwningCommand<S>, crate::Error> {
         self.as_std().over_ssh(session)
     }
 }
@@ -155,10 +156,10 @@ impl<S> OverSsh for &S
 where
     S: OverSsh,
 {
-    fn over_ssh<'session>(
+    fn over_ssh<U: Deref<Target = Session> + Clone>(
         &self,
-        session: &'session Session,
-    ) -> Result<Command<'session>, crate::Error> {
+        session: U,
+    ) -> Result<OwningCommand<U>, crate::Error> {
         <S as OverSsh>::over_ssh(self, session)
     }
 }
@@ -167,10 +168,10 @@ impl<S> OverSsh for &mut S
 where
     S: OverSsh,
 {
-    fn over_ssh<'session>(
+    fn over_ssh<U: Deref<Target = Session> + Clone>(
         &self,
-        session: &'session Session,
-    ) -> Result<Command<'session>, crate::Error> {
+        session: U,
+    ) -> Result<OwningCommand<U>, crate::Error> {
         <S as OverSsh>::over_ssh(self, session)
     }
 }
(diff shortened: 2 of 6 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/136)

masklinn:
@NobodyXu can you approve workflow runs? Also maybe comment on this early starting point.

NobodyXu:
P.S. I don't think implementing `Command<Session>` is a good idea, since it's never designed to be taken by value.

`Command<Arc<Session>>` or `Command<Rc<Session>>` is more similar to `Command<&Session>` and I'd like to see support for them.

src/child.rs
```diff
@@ -74,18 +74,20 @@ macro_rules! delegate {
...
 
     stdin: Option<ChildStdin>,
     stdout: Option<ChildStdout>,
     stderr: Option<ChildStderr>,
 }
 
-impl<'s> RemoteChild<'s> {
+pub type RemoteChild<'a> = Child<&'a Session>; 
+
+impl<T> Child<T> {
```
NobodyXu, line 88:
```suggestion
impl<S: AsRef<Session> + Clone> Child<S> {
```

src/child.rs
```diff
@@ -74,18 +74,20 @@ macro_rules! delegate {
 /// # }
 /// ```
 #[derive(Debug)]
-pub struct RemoteChild<'s> {
-    session: &'s Session,
+pub struct Child<T> {
+    session: T,
```
NobodyXu, line 78:
```suggestion
pub struct Child<S> {
    session: S,
```

src/child.rs
NobodyXu: With the `Clone` bound, you can also implement `fn session() -> S`.

src/session.rs
```diff
@@ -277,6 +277,15 @@ impl Session {
         )
     }
 
+    pub fn shared_command<'a, S: Into<Cow<'a, str>>>(self: std::sync::Arc<Self>, program: S) -> OwnedCommand<std::sync::Arc<Self>> {
```
NobodyXu, line 280: I think `shared_thread_safe_command` is a better name.
Then we can have `shared_command` for `Rc<...>`

src/session.rs
```diff
@@ -277,6 +277,15 @@ impl Session {
...
+        self.shared_raw_command(&*shell_escape::unix::escape(program.into()))
+    }
+
+    pub fn shared_raw_command<S: AsRef<OsStr>>(self: std::sync::Arc<Self>, program: S) -> OwnedCommand<std::sync::Arc<Self>> {
```
NobodyXu, line 284: I would like to also have another generic version of this that takes `this: impl AsRef<Session> + Clone`

tests/openssh.rs
```diff
@@ -1057,3 +1057,40 @@ async fn test_read_large_file_bug() {
         assert_eq!(stdout.len(), bs * count);
     }
 }
+
+
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn rc_client() {
```
NobodyXu, line 1064:
```suggestion
async fn arc_client() {
```

NobodyXu (changes requested):

src/child.rs
masklinn: Yeah I figured that afterwards but didn't have the time to update the PR.

I don't think the `AsRef` bound is useful since at the end fo the day the `Client` doesn't actually care about the `Session`, beyond the outstanding reference in the borrow case. And the `Clone` case is only useful for `Session` itself.

I guess I could bound the entire thing anyway to ensure `Child` is only used with a session-adjacent object, but it feels unnecessary?

src/session.rs
masklinn: I'm not sure `impl` can be used this way, and I'm reasonably certain it it's not supported (yet) for a `self`. It could be a free or factory function, but in that case maybe it should be a public constructor for `OwnedCommand` instead?

masklinn:
> P.S. I don't think implementing `Command<Session>` is a good idea, since it's never designed to be taken by value.

I was mostly thinking of the case where a user might want to run a single command on the session, but didn't want to bother with borrowing. However as I wrote up it doesn't work without changing the `Command` API entirely so that's moot.

src/session.rs
masklinn: [Currently valid list of self types](https://doc.rust-lang.org/reference/items/associated-items.html#methods) is `Self`, reference, box, rc, arc, and pin. So yeah no impl as an instance method.

src/session.rs
masklinn: Added that change but TBH it feels unwieldingly verbose.

src/child.rs
NobodyXu: That's true, so it's probably ok for `Child` to not constraint `S` since it is not used.

src/session.rs
NobodyXu: You can make it an associated function instead `this: impl AsRef<Session> + Clone` is valid.

masklinn: Oh I see that's what you meant, sorry I misunderstood.

Should it just replace the Arc version then? As it kinda obviates it, and if we call it "to_command" or something like that, even with an explicit `Arc` wrapper it's a lot shorter still.

NobodyXu: You could keep it as a convenient method, or remove it if you think it's unnecessary.

src/session.rs
NobodyXu: Hmmm, maybe arc_command is a better name?

src/session.rs
masklinn: I'll be honest here, I don't use the crate, I just saw the issue during jonhoo's stream and figured I had an idea which could solve the problem with not too much code duplication (or macro).

So while I do have my own sensibilities about code and APIs, I'd *much* rather go with what you, jon, and other actual users of the crate (maybe the other maintainers of openssh-sftp-client since that's where things started?) would prefer.

NobodyXu: In that case, let's keep the arc_command, AFAIK many do use it from multi-threaded async context with Arc.

masklinn: OK if it's a common use case it makes sense to have a convenient shortcut. And the associated function does look pretty good in the test.

I made the arc versions just be convenience shortcuts over the associated function as that seems simpler / clearer.

src/command.rs
```diff
@@ -207,17 +207,19 @@ where
 ///   [`ssh(1)`]: https://linux.die.net/man/1/ssh
 ///   [`env(1)`]: https://linux.die.net/man/1/env
 #[derive(Debug)]
-pub struct Command<'s> {
-    session: &'s Session,
+pub struct OwnedCommand<S> {
```
NobodyXu, line 210: Can we rename this to `CommandAnyClonable`?

NobodyXu (changes requested):

src/command.rs
masklinn: Ah sorry I had not seen your comment before doing a bit more shuffling around, squashing the commits, and writing docs.

I ended up renaming the internal command back to `Command`, and doing the aliasing on the edge in `lib.rs`, with the inner generic `Command` being exported as `OwningCommand` (seems better than `Owned`), and the `openssh::RemoteChild` and `openssh::Command` aliases being created there (`lib.rs`) rather than inside the crate guts.

I can still rename `crate::command::Command` if you prefer some other name tho.

src/session.rs
```diff
@@ -270,13 +272,85 @@ impl Session {
     ///
     /// If `program` is not an absolute path, the `PATH` will be searched in an OS-defined way on
     /// the host.
-    pub fn raw_command<S: AsRef<OsStr>>(&self, program: S) -> Command<'_> {
+    pub fn raw_command<S: AsRef<OsStr>>(&self, program: S) -> Command<&'_ Self> {
```
NobodyXu, line 275: You can forward command and raw_command to to_command and to_raw_command

src/session.rs
```diff
@@ -327,7 +401,7 @@ impl Session {
     ///
     /// # Ok(()) }
     /// ```
-    pub fn subsystem<S: AsRef<OsStr>>(&self, program: S) -> Command<'_> {
+    pub fn subsystem<S: AsRef<OsStr>>(&self, program: S) -> Command<&'_ Self> {
```
NobodyXu, line 404: Can we also have a generic version for subsystem please?

It cannot be emulated by to_command or to_raw_command

src/command.rs
```diff
@@ -167,29 +167,30 @@ impl<S> OverSsh for &mut S
...
-/// interface is almost identical to that of [`std::process::Command`].
+/// A default configuration can be generated using [`Session::command(program)`](Session::command)
+/// or [`Session::arc_command(program)`](Session::arc_command), where `program` gives a path to
+/// the program to be executed. Additional builder methods allow the configuration to be changed
+/// (for example, by adding arguments) prior to spawning. The interface is almost identical to
+/// that of [`std::process::Command`].
 ///
-/// `Command` can be reused to spawn multiple remote processes. The builder methods change the
-/// command without needing to immediately spawn the process. Similarly, you can call builder
+/// `OwnedCommand` can be reused to spawn multiple remote processes. The builder methods change
```
NobodyXu, line 187: Did you forget to update the name?

src/lib.rs
```diff
@@ -167,12 +167,16 @@ mod builder;
 pub use builder::{KnownHosts, SessionBuilder};
 
 mod command;
-pub use command::{Command, OverSsh};
+pub use command::{Command as OwningCommand, OverSsh};
```
NobodyXu, line 170: I suggest renaming the type instead of creating a type alias, it would make our doc confusing to user

NobodyXu (changes requested):

src/command.rs
masklinn: Yep, looks like I missed this one.

src/session.rs
masklinn: Interestingly no, because there is no `AsRef<T> for &T`.

src/session.rs
masklinn: Sure, should it have both the `to_` and the `arc` version, or just the `to_` version?

src/lib.rs
masklinn: Damn and here I'd thought I could avoid renaming `Command` everywhere.

src/session.rs
NobodyXu: Just to to version will be good enough.

src/session.rs
NobodyXu: Hmmm, that's unexpected

masklinn: Yep it also surprised me when I hit it. One of the reasons I'm always a little bit confused about using AsRef v Borrow v Deref.

Here AsRef might be the wrong choice because it allows "transverse" ref conversions e.g. `AsRef<OsStr> for str` and has more flexibility but I'm not sure that's needed.

`Borrow` and `Deref` are more limited but might be more suitable.

NobodyXu: In that case, maybe swap that to Deref, since reference/Arc/Rc etc all have deref implemented.

masklinn:
- moved the trait bound from AsRef to Deref
  - made `[raw_]command` go through `to_[raw_]command` since they now can
- renamed the owning `Command` to `OwningCommand` instead of just aliasing it as the crate root, for documentary clarity
- added a `to_subcommand` associated function to do the session-owning thing for subcommand features

src/session.rs
```diff
@@ -270,20 +270,89 @@ impl Session {
...
+    ///         if n == 0 {
+    ///             break;
+    ///         }
+    ///         println!("read {:?}", &buf[..n]);
+    ///     }
+    ///     Ok(())
+    /// })
+    /// # Ok(()) }
+    pub fn arc_command<'a, P: Into<Cow<'a, str>>>(
+        self: std::sync::Arc<Session>,
```
NobodyXu, line 307: I think taking a 

```suggestion
        self: &std::sync::Arc<Session>,
```

is better and more ergonomic, otherwise users would have to do `session.clone().arc_command(..)`

NobodyXu (changes requested):

NobodyXu:
BTW, you need to fix the CI before I can merge this PR.

src/session.rs
masklinn: I think the current version is better, because otherwise the arc has to be cloned internally, so if the caller just wants to create a movable command/child they can't just move the one arc they have they get additional and unnecessary refcount traffic.

masklinn:
> BTW, you need to fix the CI before I can merge this PR.

Yep but since I can't run the tests locally I can only push and wait for CI approval. I'll get on that.

src/session.rs
NobodyXu: That's reasonable.

NobodyXu:
@masklinn There's a [`run_ci_tests.sh`](https://github.com/openssh-rust/openssh/blob/master/run_ci_tests.sh) in the repository that can be used to run some of the tests.

For linting, it's just `cargo-clippy` and `cargo-fmt`.

masklinn:
> There's a [run_ci_tests.sh](https://github.com/openssh-rust/openssh/blob/master/run_ci_tests.sh) in the repository that can be used to run some of the tests.

Yeah but it requires docker, and I'm on a mac and not super willing to install docker desktop.

> For linting, it's just `cargo-clippy` and `cargo-fmt`.

Yeah I ran fmt but it changed imports I had not touched so I figured the repo had strange configuration and reverted them, and they show up as lint errors so I guess it's just that they were last fmt'd before a change in configuration or something.

masklinn:
Hopefully CI is good now, I broke one of the existing doc comments by mistake, and the new `arc_command` example had a few brainfarts.

NobodyXu:
> Yeah but it requires docker, and I'm on a mac and not super willing to install docker desktop.

You can use [OrbStack](https://orbstack.dev/), an alternative to docker desktop that claims to be faster than docker desktop and uses less energy.

I use it for docker desktop now and it indeed startup quickly.

> It's nbd as far as I'm concerned just a bit more latency but I'm not replying within seconds anyway. Unless you have to approve every CI run I can imagine that gets old.

Yes, I do need to approve every CI run.

Once this PR lands in main, any new PR from you will have CI run automatically without having me to approve.

> so I guess it's just that they were last fmt'd before a change in configuration or something.

It's probably because new lints are added to newer cargo.

tests/openssh.rs
```diff
@@ -1057,3 +1057,73 @@ async fn test_read_large_file_bug() {
...
+        let status = child.wait().await.unwrap();
+
+        // ... successfully
+        assert!(status.success());
+    }
+}
+
+#[tokio::test]
+#[cfg_attr(not(ci), ignore)]
+async fn generic_client() {
```
NobodyXu, line 1099:
```suggestion
async fn rc_client() {
```

I think `rc_client` is a better name here.

NobodyXu (changes requested):

NobodyXu:
The coverage failure is probably due to CI setup (e.g. GHA does not allow PRs from non-member to use secrets) or whatever, I can bypass that and merge the PR once you address my review.

tests/openssh.rs
masklinn: Dunno, the intent of the name was to indicate that we're testing the generic `Session::to_command` rather than the ref-specific Session::command` or the Arc-specific `Session::arc_command`, the Rc is an implementation detail of the test but I figured I might as well use that since references and arc are already exercised. Would it help if I made most of the body generic and then ran it with both Rc and Arc?

NobodyXu: It makes sense then, though I would like to rename this and the other test fn  to test_session_to_command and test_session_arc_command

masklinn: Works for me.

NobodyXu (approved):

NobodyXu (approved):

NobodyXu:
Thank you @masklinn !

NobodyXu:
Released in v0.10.1

PR #141, opened 2024-01-21
Drop MPL licensed dependency dirs-sys

pacak:
Dependabot updated `dirs` crate to version 5.0.0 here: https://github.com/openssh-rust/openssh/pull/114. Between versions 4.0 and 5.0 dirs crate included a small dependency with MPL license, there's several tickets to dirs crate about this issue closed as WONTFIX, you can find some relevant links here: https://github.com/harryfei/which-rs/pull/78

One possible solution is to pin `dirs` to some older version, but functionality used by `openssh` is simple enough to implement from scratch

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -8,10 +8,31 @@ use std::process::Stdio;
 use std::str;
 use std::{fs, io};
 
-use dirs::state_dir;
 use once_cell::sync::OnceCell;
 use tempfile::{Builder, TempDir};
 use tokio::process;
+
+#[cfg(not(windows))]
+fn state_dir() -> Option<PathBuf> {
+    fn get_absolute_path(path: OsString) -> Option<PathBuf> {
+        let path = PathBuf::from(path);
+        path.is_absolute().then_some(path)
+    }
+
+    #[allow(deprecated)]
+    if let Some(xdg) = std::env::var_os("XDG_STATE_HOME") {
+        get_absolute_path(xdg)
+    } else if let Some(home) = std::env::home_dir() {
+        Some(get_absolute_path(home.into())?.join(".local/state"))
+    } else {
+        None
+    }
+}
+
+#[cfg(windows)]
+fn state_dir() -> Option<PathBuf> {
+    None
+}
 
 /// The returned `&'static Path` can be coreced to any lifetime.
 fn get_default_control_dir<'a>() -> Result<&'a Path, Error> {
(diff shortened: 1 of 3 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/141)

src/builder.rs
```diff
@@ -8,11 +8,25 @@ use std::process::Stdio;
...
 use tempfile::{Builder, TempDir};
 use tokio::process;
 
+fn state_dir() -> Option<PathBuf> {
+    fn get_absolute_path(path: OsString) -> Option<PathBuf> {
+        let path = PathBuf::from(path);
+        path.is_absolute().then_some(path)
+    }
+
+    if let Some(xdg) = std::env::var_os("XDG_STATE_HOME") {
```
NobodyXu, line 21: note that in dirs, state_dir() always returns None for macos.

I'm not sure if we should follow their impl here, however since macos is also unix, I tend to think it's ok to use ~/.local/state

src/builder.rs
```diff
@@ -8,11 +8,25 @@ use std::process::Stdio;
...
+        get_absolute_path(xdg)
+    } else if let Some(home) = std::env::var_os("HOME") {
```
NobodyXu, line 23: On unix, if env HOME is not set, you would need to call getpwuid_r to determin home.

Luckily, std already provides it, although it's deprecated due to impl problems on windows:

```suggestion
    } else if let Some(home) = std::env::home_dir() {
```

NobodyXu (changes requested):
Thank you!

NobodyXu:
Can you also update src/changelog.rs please?

src/builder.rs
pacak: Done, I also made sure it avoids calling `home_dir()` on Windows.

src/builder.rs
```diff
@@ -157,10 +178,9 @@ impl SessionBuilder {
     /// Set the directory in which the temporary directory containing the control socket will
     /// be created.
     ///
-    /// If not set, openssh will try to use [`dirs::state_dir`] and fallback to
+    /// If not set, openssh will try to use `$XDG_STATE_HOME`, `$HOME/.local/state` and fallback to
```
NobodyXu, line 181:
```suggestion
    /// If not set, openssh will try to use `$XDG_STATE_HOME` on unix, `$HOME/.local/state` and fallback to
```

NobodyXu (changes requested):

src/builder.rs
pacak: done.

NobodyXu:
Can you rebase against main please (not merge from main though)?

NobodyXu:
I can do it for u from GH ui if you don't mind

pacak:
> I can do it for u from GH ui if you are ok with it

Too late, I already did :)

NobodyXu (approved):
Thank you!

NobodyXu:
Will cut a release later today

PR #146, opened 2024-04-25
Make `-o ControlPersist` configurable

krfricke:
The crate uses `ControlPersist=yes` per default. In practice we see the control processes around forever, and they are not reused across relaunches of the binary. So instead it would be nice to set the ControlPersist option to a custom value, so they can expire automatically after e.g. 5 minutes.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -85,7 +85,7 @@ pub struct SessionBuilder {
     server_alive_interval: Option<u64>,
     known_hosts_check: KnownHosts,
     control_dir: Option<PathBuf>,
-    control_persist: Option<String>,
+    control_persist: ControlPersist,
     clean_history_control_dir: bool,
     config_file: Option<PathBuf>,
     compression: Option<bool>,
@@ -104,7 +104,7 @@ impl Default for SessionBuilder {
             server_alive_interval: None,
             known_hosts_check: KnownHosts::Add,
             control_dir: None,
-            control_persist: None,
+            control_persist: ControlPersist::Forever,
             clean_history_control_dir: false,
             config_file: None,
             compression: None,
@@ -204,12 +204,13 @@ impl SessionBuilder {
         self
     }
 
-    /// Set the ControlPersist option (`ssh -o ControlPersist=<value>`).
+    /// Set the ControlPersist option to configure how long the controlling
+    /// ssh session should stay alive.
     ///
-    /// If not set, openssh will use "ControlPersist=yes".
+    /// Defaults to `ControlPersist::Forever`.
     ///
-    pub fn control_persist(&mut self, value: String) -> &mut Self {
-        self.control_persist = Some(value);
+    pub fn control_persist(&mut self, value: ControlPersist) -> &mut Self {
+        self.control_persist = value;
         self
     }
 
@@ -416,10 +417,7 @@ impl SessionBuilder {
             .arg("-f")
             .arg("-N")
             .arg("-o")
-            .arg(format!(
-                "ControlPersist={}",
-                self.control_persist.as_deref().unwrap_or("yes")
-            ))
+            .arg(self.control_persist.as_option())
             .arg("-o")
             .arg("BatchMode=yes")
             .arg("-o")
@@ -493,6 +491,29 @@ impl SessionBuilder {
             Err(Error::interpret_ssh_error(&output))
         } else {
             Ok(dir)
+        }
+    }
+}
+
+/// Specifies how long the controlling ssh process should stay alive.
+#[derive(Clone, Debug, Default)]
+pub enum ControlPersist {
+    /// Will stay alive indefinitely.
+    #[default]
+    Forever,
+    /// Will be closed after the initial connection is closed
+    ClosedAfterInitialConnection,
+    /// If the ssh control server has been idle for specified duration, it will
+    /// exit.
+    IdleFor(std::time::Duration),
+}
+
+impl ControlPersist {
+    fn as_option(&self) -> String {
+        match *self {
+            ControlPersist::Forever => "ControlPersist=yes".to_string(),
+            ControlPersist::ClosedAfterInitialConnection => "ControlPersist=no".to_string(),
+            ControlPersist::IdleFor(d) => format!("ControlPersist={}s", d.as_secs()),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/146)

src/builder.rs
```diff
@@ -202,6 +204,15 @@ impl SessionBuilder {
         self
     }
 
+    /// Set the ControlPersist option (`ssh -o ControlPersist=<value>`).
+    ///
+    /// If not set, openssh will use "ControlPersist=yes".
+    ///
+    pub fn control_persist(&mut self, value: String) -> &mut Self {
```
NobodyXu, line 211: Passing in an arbitrary string is bad, user could pass `yes --...` to pass any option to the ssh.

Let's create an enum to limit possible values:

```rust
#[derive(Debug, Default)]
enum ControlPersistOptions {
    #[default]
    Forever,
    // Will be closed after the initial connection is closed
    ClosedAfterInitialConnection,
    // If the ssh control server has been idle for specified duration, it will
    // exit.
    IdleFor(Duration),
}
```

NobodyXu (changes requested):

NobodyXu:
Thank you for your PR!

I have one change requested and once you resolve it, I will merge this PR and cut a new release.

krfricke:
Thanks for the quick review. It's a good point, I've updated the PR. I've called the enum `ControlPersist`, in line with the `KnownHosts` struct. I also use a simple conversion which just gives a string in seconds (e.g. 5 minutes will be `300s`, and not `5m`). Let me know if you prefer we run some maths there to give a better human readable output.

src/builder.rs
```diff
@@ -483,6 +495,29 @@ impl SessionBuilder {
     }
 }
 
+/// Specifies how long the controlling ssh process should stay alive.
+#[derive(Clone, Debug, Default)]
```
NobodyXu, line 499: Let's add `non_exhaustive` so that we can add more variants without breaking anyone:

```suggestion
#[derive(Clone, Debug, Default)]
#[non_exhaustive]
```

src/builder.rs
```diff
@@ -483,6 +495,29 @@ impl SessionBuilder {
...
+}
+
+impl ControlPersist {
+    fn as_option(&self) -> String {
+        match *self {
+            ControlPersist::Forever => "ControlPersist=yes".to_string(),
+            ControlPersist::ClosedAfterInitialConnection => "ControlPersist=no".to_string(),
+            ControlPersist::IdleFor(d) => format!("ControlPersist={}s", d.as_secs()),
+        }
+    }
```
NobodyXu, line 518: Could use `Cow` to avoid unnecessary `String` creation:

```suggestion
    fn as_option(&self) -> String {
        match self {
            ControlPersist::Forever => Cow::Borrowed("ControlPersist=yes"),
            ControlPersist::ClosedAfterInitialConnection => Cow::Borrowed("ControlPersist=no"),
            ControlPersist::IdleFor(d) => format!("ControlPersist={}s", d.as_secs()).into(),
        }
    }
```

src/builder.rs
```diff
@@ -483,6 +495,29 @@ impl SessionBuilder {
...
+/// Specifies how long the controlling ssh process should stay alive.
+#[derive(Clone, Debug, Default)]
+pub enum ControlPersist {
+    /// Will stay alive indefinitely.
+    #[default]
+    Forever,
+    /// Will be closed after the initial connection is closed
+    ClosedAfterInitialConnection,
+    /// If the ssh control server has been idle for specified duration, it will
+    /// exit.
```
NobodyXu, line 507:
```suggestion
    /// If the ssh control server has been idle for specified duration, it will
    /// exit.
    ///
    /// If you specify a value less than 1 second, then it would be treated as expiring
    /// immediately if the control server is idle.
```

NobodyXu (changes requested):
Just a little bit minor suggestions:

src/builder.rs
krfricke: According to the ssh_config docs, passing `0` (or `0s`) is the same as passing `yes`. I've tried this manually just now and the session sticks around forever.

If we want the behavior you describe, we can either set set a minimum of 1 second.

I think putting a requirement of 1 second is preferable rather than having it stick around forever -- if someone uses `IdleFor` they will expect the session to eventually go away.

```suggestion
    /// If the ssh control server has been idle for specified duration, it will
    /// exit.
    ///
    /// If you specify a value less than 1 second, it will be rounded up to
    /// 1 second (the minimum possible time)
```

krfricke:
Updated - please review my usage of `Cow`. I believe with `deref()` it should avoid allocations in the non-IdleFor cases.

src/builder.rs
```diff
@@ -405,7 +418,7 @@ impl SessionBuilder {
             .arg("-f")
             .arg("-N")
             .arg("-o")
-            .arg("ControlPersist=yes")
+            .arg(self.control_persist.as_option().deref())
```
krfricke, line 421: Without the `deref()` it complains with `the trait AsRef<OsStr> is not implemented for Cow<'_, str>`

src/builder.rs
NobodyXu: Yeah that sounds more reasonable.

We should also warn user that if they pass a duration set to 0, then it would be the same as Forever.

NobodyXu: Or, we could replace Duration with NonZeroUsize, representing number of seconds, which is less confusing.

What do you think?

src/builder.rs
NobodyXu: Yeah, this is ok

src/builder.rs
NobodyXu: cc @krfricke just in case you missed my comment

NobodyXu (changes requested):
BTW, can you update `src/changelog.rs` for the changelog?

src/builder.rs
krfricke: I think either is fine (Duration that is rounded to minimum of 1 second, or NonZeroUsize). Both are usually constructed from seconds as a unit, and converting between them is also pretty straightforward.

But I do like the semantics of NonZeroUsize, as it avoids the rounding and the millisecond precision of duration is ignored anyway. So I'll update the PR to use that instead

krfricke:
All updated and moved to NonZeroUsize. Thanks for your help

NobodyXu (approved):
Thank you!

PR #155, opened 2024-08-10
Remove dep tokio-pipe

NobodyXu:
also remove deprecated functions, and replace `From<tokio::proces::Child*>`
with `TryFrom<tokio::proces::Child*>`, since the converison is falliable.

Also remove `IntoRawFd` for `Child*` since the conversion is falliable.


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/155)

PR #179, opened 2025-12-03
fix: encoding of IPv6 addresses in `impl From<SocketAddr> for Socket`

paolobarbolini:
This makes the `impl From<SocketAddr> for Socket<'static>` implementation wrap IPv6 addresses in square brackets, so that an address like an address like `2606:4700:4700::1111` port `123` gets encoded as `[2606:4700:4700::1111]:123` instead of `2606:4700:4700::1111:123` as it does now.

This is also what the std `Display` implementation for `SocketAddr` does: https://doc.rust-lang.org/1.91.1/src/core/net/socket_addr.rs.html#678

### DIFF
--- a/src/port_forwarding.rs
+++ b/src/port_forwarding.rs
@@ -6,7 +6,7 @@ use std::ffi::OsStr;
 
 use std::borrow::Cow;
 use std::fmt;
-use std::net::{self, SocketAddr};
+use std::net::{self, IpAddr, SocketAddr};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


jonhoo:
This change is [Reviewable](https://reviewable.io/reviews/openssh-rust/openssh/179)

NobodyXu (approved):
Thank you! Will cit a release now

paolobarbolini:
That was quick! Thank you very much :smiley:

